PR #5, opened 2021-05-20
feature: support hydration of nested resources that are paginated

rocktavious:


client_test.go
```diff
@@ -86,12 +85,15 @@ func TestClientQuery(t *testing.T) {
 	autopilot.Equals(t, "1234", q.Account.Id)
 }
 
+/*
+These tests don't work very well with our autopilot endpoint stuff because they need to make recursive calls
+We need to figure out a better way to handle mapping different payloads to different requests
+
```
rocktavious, line 91: I tried writing some tests for the pagination stuff but its not working correctly and i think its because of how i'm mocking out the opslevel API.  I will have to come back later and figure out a better solution for mocking the opslevel API server for recursive pagination calls.

rocktavious:
I'm pretty confident the logic for this works but the OpsLevel API is currently not returning valid nested pagination responses so i cannot 100% validate nor can i write tests for this without refactoring how i mock the OpsLevel API server.

PR #9, opened 2021-07-28
implement CRUD for filters

rocktavious:


filters.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	if err := client.Query(&q, nil); err != nil {
+		return q.Account.Filters.Nodes, err
+	}
+	if err := q.Account.Filters.Hydrate(client); err != nil {
+		return q.Account.Filters.Nodes, err
+	}
+	return q.Account.Filters.Nodes, nil
+}
+
```
rocktavious, line 111: Missing Get by ID and Delete

rocktavious: delete implement - need to wait for Get by ID coming later in the API

PR #37, opened 2022-01-12
Fix file name typo (integraion -> integration)

eapache:
@kenrose @rocktavious :)

On an ipad, so I haven’t actually `go test`ed this unfortunately.


rocktavious:
Thanks @eapache for the contribution!

PR #38, opened 2022-01-25
fix: export option type

george-e-shaw-iv:
_This shouldn't be a breaking change, it doesn't update or remove from any of the public-facing API._

If this type isn't exported it makes it pretty difficult to write types that wrap around opslevel.Client. Right now if you want to support setting the URL/page size on a wrapped client with functional options, for example, you have to do one of the following hacky options:

1.
```go
switch {
case do.url != nil && do.pageSize != nil:
	// both are non-nil, apply both.
	return opslevel.NewClient(apiToken, opslevel.SetContext(ctx), opslevel.SetURL(*do.url), opslevel.SetPageSize(*do.pageSize))
case do.url != nil:
	// url is non-nil, only apply it.
	return opslevel.NewClient(apiToken, opslevel.SetContext(ctx), opslevel.SetURL(*do.url))
case do.pageSize != nil:
	// pageSize is non-nil, only apply it.
	return opslevel.NewClient(apiToken, opslevel.SetContext(ctx), opslevel.SetPageSize(*do.pageSize))
default:
	// neither were provided, just return a client with the context set.
	return opslevel.NewClient(apiToken, opslevel.SetContext(ctx))
}
```
(extremely verbose and gets pretty clunky pretty quickly, requiring 2^n cases where n is the number of `opslevel.option`s you want to support)

2.
```go
urlOpt := opslevel.SetContext(ctx)
if do.url != nil {
    urlOpt = opslevel.SetURL(*do.url)
}

pageOpt := opslevel.SetContext(ctx)
if do.pageSize != nil {
    pageOpt = opslevel.SetPageSize(*do.pageSize)
}
...
return opslevel.NewClient(apiToken, opslevel.SetContext(ctx), urlOpt, pageOpt, ...)
```
(this works because you can technically pass `opslevel.SetContext` more than once to `opslevel.NewClient`)

3.
```go
return opslevel.NewClient(apiToken, func(c *ClientSettings) {
   if do.url != nil {
      opslevel.SetURL(*do.url)(c)
   }
   ....
})
```
(works because go will implicitly assert anonymous functions even if the type is unexported, will not work with an explicit custom type)

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,3 +167,68 @@ for _, alias := range m.Payload.Aliases {
 	fmt.Println(alias)
 }
 ```
+
+### Wrapping the Client
+
+Wrapping the client can be useful when you want to override default behavior, such as always setting
+context or disallowing (to the best of Go's ability) access to specific receiver functions on
+`opslevel.Client`. Here is an example of wrapping the client to always require context to be passed
+while also maintaining the ability to pass in default options, with the ability to extend your own
+set of options if need-be:
+
+```go
+type Client struct{
+	*opslevel.Client
+
+	do opsLevelDefaultOptions
+	customOption string
+}
+
+func NewClient(ctx context.Context, apiToken string, options ...Option) *Client {
+	var c Client
+
+	for i := range options{
+		options[i](&c)
+	}
+	c.Client = opslevel.NewClient(apiToken, c.do.opsLevelOptions(ctx)...)
+
+	return &c, nil
+}
+
+type opsLevelDefaultOptions struct{
+	url *string
+	pageSize *int
+}
+
+func (o *opsLevelDefaultOptions) opsLevelOptions(ctx context.Context) []opslevel.Option {
+	opts := []opslevel.Option{opslevel.SetContext(ctx)} // Always set the context.
+
+	if o.url != nil {
+		ops = append(opts, opslevel.SetURL(*o.url))
+	}
+	if o.pageSize != nil {
+		ops = append(opts, opslevel.SetPageSize(*o.pageSize))
+	}
+}
+
+// Option is our own functional option type for our own custom client.
+type Option func(*Client)
+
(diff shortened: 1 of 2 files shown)


rocktavious:
@george-e-shaw-iv - While i feel like option 2 is probably the closest to idiomatic go can I see the pain here.

Would you mind adding a section in the readme under `Advanced Usage` for `configuring the client` which gives an example of how you are wrapping the client with this change in-place?

That will help direct others on the correct way to wrap it now that this will be exported.

george-e-shaw-iv:
> @george-e-shaw-iv - While i feel like option 2 is probably the closest to idiomatic go can I see the pain here.
> 
> Would you mind adding a section in the readme under `Advanced Usage` for `configuring the client` which gives an example of how you are wrapping the client with this change in-place?
> 
> That will help direct others on the correct way to wrap it now that this will be exported.

added in newest commit

rocktavious:
@george-e-shaw-iv - after seeing the e-mail the only thing i'm considering now is should the functional option's type be called just `Option` or should we use something like `ClientOption`. 

Currently the example makes it slightly hard to reason about since there are 2 `Option` and `opslevel.Option` and I found myself now following the e-mail clearly sometimes.

Thoughts on renaming it now before its public and harder to change?

george-e-shaw-iv:
That's up to y'all really. I think it's fine because the types are basically "namespaced" by packages. I understand how it could be confusing though. If you want me to change your internal type on this branch from `Option` to `ClientOption`, I can do that, or you can push a commit to this branch too. @rocktavious

rocktavious:
Ok @george-e-shaw-iv - I want to get 1 more set of eyes on this just to validate its ok being `Option`. I will circle back around tomorrow to merge if that second pair of eyes agrees its good.

Thanks for the contribution.

PR #40, opened 2022-02-01
Move examples to subdir

eochoa260:
Moved each example into its own file and updated readme.


examples/client.go
```diff
@@ -0,0 +1,63 @@
+/*
+Wrapping the client can be useful when you want to override default behavior, such as always setting
```
eochoa260, line 2: I wasn't sure if we wanted to retain this comment in the root readme, create a new readme in the examples dir?

examples/mutation.go
```diff
@@ -0,0 +1,19 @@
+var mutation struct {
```
eochoa260, line 1: Should I be importing packages in these files?  I just copied what was in the readme for now.

rocktavious: I think this would be helpful context to have in the example - i would add it

examples/client.go
rocktavious: lets go with this for now - if someone is digging through example code having info inline like this is probably what they'd expect

rocktavious (commented):
Lets make the imports and package name changes and call it good.

rocktavious (approved):
LGTM

PR #45, opened 2022-04-01
Adding Group get and list

eochoa260:


group.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := client.Query(q, v); err != nil {
+		return conn.Nodes, err
+	}
+	if err := conn.Hydrate(client); err != nil {
+		return conn.Nodes, err
+	}
+	return conn.Nodes, nil
+}
+
+func (conn *GroupConnection) Hydrate(client *Client) error {
```
eochoa260, line 87: Wasn't sure if I should copy this over as is, leave it out, or fix it since I think this still has an open ticket because it only runs Hydrate function on everything after the first page?

group_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func TestListGroupsWithParent(t *testing.T) {
+	// Arrange
+	client := ATestClientAlt(t, "group/list", "group/list_with_parent")
+	// Act
+	result, err := client.ListGroupsWithParent("test_group_1")
+	// Assert
+	autopilot.Ok(t, err)
+	autopilot.Equals(t, 2, len(result))
```
eochoa260, line 52: Copied this over from service tests but this is the only part that doesn't make sense to me.  I know we're mocking the request and response but I would expect the response to only have 1 item in the result.  I understand if it's for simplicity though so we don't have to have a different response per request.

group.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	v := client.InitialPageVariables()
+	return q.Account.Groups.Query(client, &q, v)
+}
+
+func (client *Client) ListGroupsWithParent(parent string) ([]Group, error) {
+	var q struct {
+		Account struct {
+			Groups GroupConnection `graphql:"groups(parentAlias: $parent, after: $after, first: $first)"`
```
rocktavious, line 125: I do not see this on the public API for groups - lets remove this function

rocktavious (changes requested):
Overall looks good just a function that was implemented that needs to be removed

group.go
eochoa260: Ah ok, how do I check that?  I saw that services had a bunch of listwith functions so figured I had to implement the same here.

group.go
eochoa260: Ok, first page *should* get hydrated now.

eochoa260:
> 

Ok, removed the listwithparent function as discussed and took a pass at updating the Hydrate function.

group.go
```diff
@@ -0,0 +1,131 @@
...
+	return conn.Nodes, nil
+}
+
+func (conn *GroupConnection) Hydrate(client *Client) error {
+	var q struct {
+		Account struct {
+			Groups GroupConnection `graphql:"groups(after: $after, first: $first)"`
+		}
+	}
+	v := PayloadVariables{
```
eochoa260, line 96: Is this needed?  It seems to me that the initial query is already getting the first page when calling initialPageVariables()?

group.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (client *Client) GetGroupWithAlias(alias string) (*Group, error) {
+	var q struct {
+		Account struct {
+			Group Group `graphql:"group(alias: $group)"`
+		}
+	}
+	v := PayloadVariables{
+		"group": graphql.String(alias),
```
rocktavious, line 72: is the `graphql.String` wrapping needed here?  If so then its fine - if we can do without then that would be better.

group.go
rocktavious: well we'd have to pass the PayloadVariables through the Hydrate functions which is probably not what we want.

rocktavious (approved):
LGTM - Responded to your 1 and Add 1 comment of my own

group.go
eochoa260: Ah ok, that makes sense since it's used later when calling for the next page.

group.go
eochoa260: Not sure, this is how it was for service.  Is it possible this is here in case `alias="123"` and it's making sure that graphql reads it as a string instead of an integer?

PR #47, opened 2022-04-11
Add SubResource method to Group for DescendantTeams

rocktavious:


group.go
```diff
@@ -143,6 +145,38 @@ func (client *Client) ListGroups() ([]Group, error) {
...
+				} `graphql:"descendantTeams(after: $after, first: $first)"`
+			} `graphql:"group(id: $group)"`
+		}
+	}
+	if g.Id == nil {
+		return nil, fmt.Errorf("failed to request 'DescendantTeams' for group because there is not a valid id '%s'", g.Id)
+	}
+	v := PayloadVariables{
+		"group": g.Id,
+		"first": client.pageSize,
```
rocktavious, line 164: hmm we need an "after" var here i think

group.go
```diff
@@ -143,6 +145,38 @@ func (client *Client) ListGroups() ([]Group, error) {
...
+	}
+	output := []Team{}
+	if err := client.Query(&q, v); err != nil {
+		return nil, err
+	}
+	for q.Account.Group.DescendantTeams.PageInfo.HasNextPage {
```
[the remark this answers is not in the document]
rocktavious: hmm maybe - i'll have to test it out.  This might be the same mistake I was making before - HAHA

group.go
```diff
@@ -30,11 +32,11 @@ type Group struct {
 	// DescendantRepositories RepositoryConnection `json:"descendantRepositories,omitempty"`
 	// DescendantServices ServiceConnection `json:"descendantServices,omitempty"`
 	// DescendantSubgroups    SubgroupConnection   `json:"descendantSubgroups,omitempty"`
-	// DescendantTeams        TeamConnection       `json:"descendantTeams,omitempty"`
+	// DescendantTeamsConnection Connection `json:"descendantTeams,omitempty"`
```
eochoa260, line 35: Was having a hard time with these since it was adding a query parameter that doesn't exist, e.g. `descendantTeamsConnection{}`.  Also, I'm not fully understanding why these would need to be in the struct so might need a quick sync when @rocktavious is back.

group.go
```diff
@@ -134,6 +136,176 @@ func (client *Client) ListGroups() ([]Group, error) {
...
+			Group struct {
+				DescendantTeams struct {
+					Nodes    []Team
+					PageInfo PageInfo
+				} `graphql:"descendantTeams(after: $after, first: $first)"`
+			} `graphql:"group(id: $group)"`
+		}
+	}
+	if g.Id == nil {
+		return nil, fmt.Errorf("failed to request 'DescendantTeams' for group because there is not a valid id '%s'", g.Id)
```
rocktavious, line 151: I know i probably wrote this but reading it back upon review - it will be a confusing error for end users to read.

Do you have any ideas on how to make it clearer?
Maybe
```
fmt.Error("unable to get 'Teams' for group with missing id")
```

That or maybe we should just return an empty team array and no error.

```
output := []Team{}
if g.Id == nil {
    return output, nil
}
```

group_test.go
```diff
@@ -50,6 +61,54 @@ func TestDeleteGroupWithAlias(t *testing.T) {
...
+	group, err := client1.GetGroupWithAlias("test_group_1")
+	result, err := group.DescendantTeams(client2)
+	// Assert
+	autopilot.Ok(t, err)
+	autopilot.Equals(t, "platform", result[0].Alias)
+}
+
+func TestDescendantRepositories(t *testing.T) {
+	// Arrange
+	client1 := getGroupWithAliasTestClient(t)
```
rocktavious, line 78: This is incredibly inventive!

testdata/fixtures/group/descendant_repositories_request.json
```diff
@@ -0,0 +1,8 @@
+{
+    "query": "query($after:String!$first:Int!$group:ID!){account{group(id: $group){descendantRepositories(after: $after, first: $first){nodes{archivedAt,createdOn,defaultAlias,defaultBranch,description,forked,htmlUrl,id,languages{name,usage},lastOwnerChangedAt,name,organization,owner{alias,id},private,repoKey,services{edges{atRoot,node{id},paths{href,path},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id}}},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor},totalCount},tags{nodes{id,key,value},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor},totalCount},tier{alias,description,id,index,name},type,url,visible},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor}}}}}",
+    "variables": {
+      "first": 100,
+      "group": "Z2lkOi8vb3BzbGV2ZWwvTmFtZXNwYWNlczo6R3JvdXAvMTI",
+      "after": ""
```
rocktavious, line 6: is this the right value?

testdata/fixtures/group/descendant_repositories_response.json
```diff
@@ -0,0 +1,17 @@
+{
+    "data": {
+      "account": {
+        "group": {
+          "descendantRepositories": {
+            "nodes": [],
```
rocktavious, line 6: this should have data in it

group.go
```diff
@@ -134,6 +136,176 @@ func (client *Client) ListGroups() ([]Group, error) {
...
+	if err := client.Query(&q, v); err != nil {
+		return nil, err
+	}
+	output = append(output, q.Account.Group.DescendantServices.Nodes...)
+	for q.Account.Group.DescendantServices.PageInfo.HasNextPage {
+		v["after"] = q.Account.Group.DescendantServices.PageInfo.End
+		if err := client.Query(&q, v); err != nil {
+			return nil, err
+		}
+		output = append(output, q.Account.Group.DescendantServices.Nodes...)
```
rocktavious, line 236: We may need to call Hydrate on every service object because it has sub resources with pagination.

We may want to consider changing this to return `ServiceId` rather then a `Service` object which then forces the user to make additional API calls to turn a ServiceId into a full Service object with all the pagination

group.go
```diff
@@ -134,6 +136,176 @@ func (client *Client) ListGroups() ([]Group, error) {
...
+	if err := client.Query(&q, v); err != nil {
+		return nil, err
+	}
+	output = append(output, q.Account.Group.DescendantTeams.Nodes...)
+	for q.Account.Group.DescendantTeams.PageInfo.HasNextPage {
+		v["after"] = q.Account.Group.DescendantTeams.PageInfo.End
+		if err := client.Query(&q, v); err != nil {
+			return nil, err
+		}
+		output = append(output, q.Account.Group.DescendantTeams.Nodes...)
```
rocktavious, line 168: Do teams have sub resources that have pagination - i think they do.  We might need to convert all these to only return `TeamID` types to not mess with the recursive pagination.

testdata/fixtures/group/members_request.json
```diff
@@ -0,0 +1,8 @@
+{
+    "query": "query($after:String!$first:Int!$group:ID!){account{group(id: $group){members(after: $after, first: $first){nodes{email,htmlUrl,id,name,role},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor}}}}}",
+    "variables": {
+      "first": 100,
+      "group": "Z2lkOi8vb3BzbGV2ZWwvTmFtZXNwYWNlczo6R3JvdXAvMTI",
+      "after": ""
```
rocktavious, line 6: i think this is missing a value

rocktavious (commented):
Overall its coming along but I think we should have a discussion around the FullObject for IDObject because of the recursive pagination

group.go
eochoa260: Yeah, I like the idea of letting the user know that there's no Id on the group. Curious what scenario would cause that to happen, would it be that they passed in an invalid group?  If that's the case then I think we should let them know that in the error.

testdata/fixtures/group/descendant_repositories_request.json
eochoa260: Nope, looks like I forgot to go back and update this.

group.go
eochoa260: Do you mean as a user, they would run the subcommand and get something similar to a list, e.g. Name, Alias, ID?  Then they would run `get service <ID>` after to get more info on whichever service they're interested in?

group.go
eochoa260: Yeah, I think the members field.

group.go
rocktavious: @eochoa260 it could be as simple as just
```
group := opslevel.Group{}
team := group.DescendantTeams()
```
but it could also just be not error checking when
```
group, _ := client.GetGroup("group-does-not-exist")
team := group.DescendantTeams()
```

group.go
rocktavious: ya exactly - that way we avoid infinite recursion

group.go
rocktavious: Then ya another case for TeamId

rocktavious:
LGTM

eochoa260 (approved):

PR #48, opened 2022-04-12
fix update group method

rocktavious:
The method did not work as built because the `groupUpdate` mutation uses that new format for taking that identifier input.

@eochoa260 can you fix the test fixtures in this branch?

### DIFF
--- a/group.go
+++ b/group.go
@@ -164,9 +164,7 @@ func (client *Client) DeleteGroupWithAlias(alias string) error {
 		Payload ResourceDeletePayload `graphql:"groupDelete(resource: $input)"`
 	}
 	v := PayloadVariables{
-		"input": IdentifierInput{
-			Alias: graphql.String(alias),
-		},
+		"input": *NewIdentifier(alias),
 	}
 	if err := client.Mutate(&m, v); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


eochoa260:
> The method did not work as built because the `groupUpdate` mutation uses that new format for taking that identifier input.
> 
> @eochoa260 can you fix the test fixtures in this branch?

Ah yeah, I forgot to update it to use IdentifierInput.  And definitely, I'll update the fixtures after my meetings this morning.

rocktavious:
Outstanding things we need to manually test:

- can we remove graphql.String from IdentifierInput ?
- Should GroupInput.Members be nullable?
- Should GroupInput.Teams be nullable?

group.go
```diff
@@ -169,15 +161,11 @@ func (client *Client) UpdateGroup(input GroupUpdateInput) (*Group, error) {
...
-			Id     graphql.ID       `graphql:"deletedId"`
-			Alias  graphql.String   `graphql:"deletedAlias"`
-			Errors []OpsLevelErrors `graphql:"errors"`
-		} `graphql:"groupDelete(input: $input)"`
+		Payload ResourceDeletePayload `graphql:"groupDelete(resource: $input)"`
 	}
 	v := PayloadVariables{
-		"input": ResourceDeletePayload{
-			Alias: alias,
+		"input": IdentifierInput{
```
eochoa260, line 167: Should NewIdentifier func be used here too?

rocktavious: fixed

PR #49, opened 2022-04-20
Add support for new check types

rocktavious:


check.go
```diff
@@ -13,24 +13,33 @@ type CheckOwner struct {
...
-	Type        CheckType  `graphql:"type"`
-
-	CustomEventCheckFragment      `graphql:"... on CustomEventCheck"`
-	ManualCheckFragment           `graphql:"... on ManualCheck"`
-	RepositoryFileCheckFragment   `graphql:"... on RepositoryFileCheck"`
-	RepositorySearchCheckFragment `graphql:"... on RepositorySearchCheck"`
-	ServicePropertyCheckFragment  `graphql:"... on ServicePropertyCheck"`
-	TagDefinedCheckFragment       `graphql:"... on TagDefinedCheck"`
-	ToolUsageCheckFragment        `graphql:"... on ToolUsageCheck"`
+	Category    Category     `graphql:"category"`
```
eochoa260, line 16: Is there a reason to not include campaign here?

rocktavious: We are not sure we are building support into the tools for that yet.

Also if we did we'd do it as a separate initiative

eochoa260 (approved):
LGTM

PR #51, opened 2022-04-22
Update fixtures after adding aliases to ServiceId

eochoa260:


testdata/fixtures/group/descendant_services_response.json
```diff
@@ -5,7 +5,11 @@
         "descendantServices": {
           "nodes": [
             {
-              "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8yNjE5"
+              "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8yNjE5",
+              "aliases": [
```
eochoa260, line 9: Just realized that I have to do this for all the responses corresponding to all the request queries that I updated. :/

service.go
```diff
@@ -7,7 +7,8 @@ import (
 )
 
 type ServiceId struct {
-	Id graphql.ID `json:"id"`
+	Id      graphql.ID `json:"id"`
+	Aliases []string   `json:"aliases,omitempty"`
 }
 
 type Service struct {
```
rocktavious, line 14: We need to remove this from the `Service` struct since it inherits from `ServiceId`

eochoa260: Great catch, I didn't see it was already using ServiceId.  I'll update it now.

PR #57, opened 2022-05-17
Eo/add child teams

eochoa260:


group.go
```diff
@@ -16,6 +16,7 @@ type Group struct {
 	DescendantServicesConnection     Connection `json:"-" graphql:"descendantServices"`
 	DescendantSubgroupsConnection    Connection `json:"-" graphql:"descendantSubgroups"`
 	DescendantTeamsConnection        Connection `json:"-" graphql:"descendantTeams"`
+	ChildTeamsConnection             Connection `json:"-" graphql:"childTeams"`
```
rocktavious, line 19: With the way we wrote our sub queries - we no longer need this.  This is not used anywhere in the code it seems.  I suspect the other connection fields are also no longer used?

eochoa260: Yeah, since it gets the page info from the query within the ChildTeams function?

eochoa260: Removed all the *Connection fields

rocktavious (approved):
LGTM

PR #63, opened 2022-06-21
Add automatic retry client

rocktavious:
While digging around in gitlab's graphql library i found this.  This should fix our customer reported problems with no retry.

The rules in the https://github.com/hashicorp/go-retryablehttp default retry policy are:

This will retry for 429 - respecting the `Retry-After` header time
This will retry for any 5xx - except for 501

This will not retry if
  - too many redirects
  - TLS cert verification failure
  
 We should do some extensive user testing of this before its released.


eochoa260 (approved):
LGTM

DougEdey:
I'm wondering if we should allow this to be configurable for the number of retries, 10 retries with a 10s timeout feels too high, I'd recommend 3-5 to start with and increase it manually if we find there's issues

rocktavious:
@DougEdey - Both are configurable when you create an instance of the client in code.  Golang has a paradigm of using "FunctionalOptions" where you pass a function that configures the struct.

To do this it would look something like this:
```
opslevel.NewClient("XXX", opslevel.SetTimeout(5), opslevel.SetMaxRetries(5))
```

rocktavious:
This was merged into https://github.com/OpsLevel/opslevel-go/pull/68

PR #67, opened 2022-07-11
Fix display name when connecting svc to repo

rcrowe:
It seems more common to expect the format to be `{org}/{repo}` as that is common with most hosted VCS suppliers.


rocktavious:
😂 thanks @rcrowe great catch!

rocktavious (approved):
LGTM

PR #72, opened 2022-07-22
Add the API Documents fields and methods for eventual use in Terraform

rocktavious:


client.go
```diff
@@ -22,7 +22,7 @@ type Option func(*ClientSettings)
 
 func newClientSettings(options ...Option) *ClientSettings {
 	settings := &ClientSettings{
-		url:     "https://app.opslevel.com/",
+		url:     "https://app.opslevel.com",
```
rocktavious, line 25: This fixes a bug with `go generate`

clientGQL.go
```diff
@@ -62,7 +62,6 @@ func NewGQLClient(options ...Option) *Client {
 	} else {
 		url = fmt.Sprintf("%s/graphql", settings.url)
 	}
-	fmt.Println(url)
```
rocktavious, line 65: This is incorrectly merged into main - cleanup 🧹

document.go
```diff
@@ -0,0 +1,38 @@
...
+type ServiceDocument struct {
+	Id         graphql.ID            `graphql:"id" json:"id"`
+	HtmlURL    string                `graphql:"htmlUrl" json:"htmUrl,omitempty"`
+	Source     ServiceDocumentSource `graphql:"source" json:"source"`
+	Timestamps Timestamps            `graphql:"timestamps" json:"timestamps"`
+}
+
+type ServiceDocumentContent struct {
+	ServiceDocument
+	Content string `graphql:"content" json:"content,omitempty"`
```
rocktavious, line 19: I pulled this out to a dedicated struct so that we are not always pulling back the full contents of the document.  Its currently unused but her incase we need to use it downstream in the CLI or Terraform

DougEdey (approved):
Do we have ways to add tests in to confirm we're deserializing the data correctly? I.e. the Go struct format can deserialize the JSON Response correctly

eochoa260 (approved):
LGTM

rocktavious:
@DougEdey - good call - I realize now i'm only testing the query side.  I'll put some data into the existing Get and List tests fixtures.

DougEdey:
I'm a bit surprised that codecov is down, but thanks for adding the tests!

DougEdey (approved):

rocktavious:
@DougEdey its because I didn't add a test for `ServiceApiDocSettingsUpdate` - Thanks for the callout i'll do this.

rocktavious:
@DougEdey and @eochoa260 - I made some material changes to `ServiceApiDocSettingsUpdate` because of the new tests i wrote which sus'd out some actual bugs.

clientGQL_test.go
```diff
@@ -71,33 +71,33 @@ func FixtureQueryValidation(t *testing.T, fixture string) autopilot.RequestValid
 }
 
 func ATestClient(t *testing.T, endpoint string) *ol.Client {
-	return ol.NewGQLClient(ol.SetAPIToken("x"), ol.SetURL(autopilot.RegisterEndpoint(fmt.Sprintf("/LOCAL_TESTING/%s", endpoint),
+	return ol.NewGQLClient(ol.SetAPIToken("x"), ol.SetMaxRetries(0), ol.SetURL(autopilot.RegisterEndpoint(fmt.Sprintf("/LOCAL_TESTING/%s", endpoint),
```
DougEdey, line 74: I figured this would have been a compilation error without the maxRetries?

service.go
```diff
@@ -12,19 +13,23 @@ type ServiceId struct {
 }
 
 type Service struct {
-	Description string `json:"description,omitempty"`
-	Framework   string `json:"framework,omitempty"`
-	HtmlURL     string `json:"htmlUrl"`
+	ApiDocumentPath string `json:apiDocumentPath,omitempty`
```
DougEdey, line 16: Should this have double quotes like the others?

```suggestion
	ApiDocumentPath string `json:"apiDocumentPath,omitempty"`
```

DougEdey (approved):

clientGQL_test.go
rocktavious: Hmm not sure what you mean.  What i'm doing his is removing the default amount of retries on the test client so that It only tries 1 time instead of 5 - which hugely slows down when tests are fail.  This makes them snappy again when you have a failing test.

service.go
rocktavious: Great Catch!

rocktavious:
@DougEdey Looks like coverage has gone up now that i've written the new tests! 🎉

clientGQL_test.go
DougEdey: This is probably a go method signature thing,  you've added a new positional parameter to `NewGQLClient` call based on my reading here.

DougEdey:
> @DougEdey Looks like coverage has gone up now that i've written the new tests! 🎉

Still less than 100% coverage of code added :P Just above the existing average!

clientGQL_test.go
rocktavious: Ahh - yes this is the "spread" operator on parameters - AKA [Variadic Function Parameters](https://yourbasic.org/golang/three-dots-ellipsis/)

The signature of NewGQLClient use `...` on the `Option` type which is to take in any N arguments such that variable `options` is effectively of type `[]Option`.
```
func NewGQLClient(options ...Option) *Client {
```

This is also using the pattern dubbed [Functional Options](https://golang.cafe/blog/golang-functional-options-pattern.html#:~:text=The%20Go%20(Golang)%20Functiona%20Options,and%20initialisation%20of%20your%20struct.)

and then latter in the function we do this to apply all the "Functional Options" to the client before returning it.

```
        settings := &ClientSettings{
	for _, opt := range options {
		opt(settings)
	}
```

DougEdey: Ah thanks for the clarification!

eochoa260 (approved):

PR #74, opened 2022-07-27
Eo/3243 ctm issue

eochoa260:


check.go
```diff
@@ -21,7 +21,7 @@ type Check struct {
 	Id          graphql.ID   `graphql:"id"`
 	Level       Level        `graphql:"level"`
 	Name        string       `graphql:"name"`
-	Notes       string       `graphql:"notes"`
+	Notes       *string      `graphql:"notes"`
```
rocktavious, line 24: This is not needed as this is on the "response" side.  The passing of null is only needed on the "request variables" side IE input

check.go
```diff
@@ -133,7 +133,7 @@ type CheckUpdateInput struct {
 	Level    *graphql.ID   `json:"levelId,omitempty"`
 	Owner    *graphql.ID   `json:"ownerId,omitempty"`
 	Filter   *graphql.ID   `json:"filterId,omitempty"`
-	Notes    string        `json:"notes,omitempty"`
+	Notes    *string       `json:"notes"`
```
rocktavious, line 136: this doesn't actually need to be a pointer we just need to remove the `omitempty` to the "input" on the update mutation passes and empty string - the `omitempty` is droping the field "notes" from the variables on the mutation.

check.go
```diff
@@ -113,7 +113,7 @@ type CheckCreateInput struct {
 	Level    graphql.ID    `json:"levelId"`
 	Owner    *graphql.ID   `json:"ownerId,omitempty"`
 	Filter   *graphql.ID   `json:"filterId,omitempty"`
-	Notes    string        `json:"notes,omitempty"`
+	Notes    *string       `json:"notes"`
```
rocktavious, line 116: no need for a pointer here - and we should keep the `omitemtpy` since this is on the "create" side.  The "update" is what the problem is.

rocktavious (changes requested):
Let me know if this all doesn't make sense - we can walk through it on a call.  Its a tricky thing so I want to make sure you fully understand the root of the problem.

eochoa260:
@rocktavious ok, I tested with `""`, `null`, and text and the update worked in all cases.  I tried creating a test case on the opslevel-go side but couldn't get it to work.  I couldn't set Notes field to nil since it said that it wasn't a valid string value.

rocktavious (approved):
👍

PR #75, opened 2022-08-05
Add has documentation check

jasonopslevel:
This PR has a few commits:

- ran go fmt and there were a few deltas, so wanted to start with a good baseline
- it seems as though there were some manual changes added to enum.go, as well as being behind in the types. So, added lots of types + also fixed older types.
- Added a section to the CONTRIBUTING.md just to point out that `gen.go` exists and how we should use it
*** Let me know if you want a separate PR for the above commits
- Added the HasDocumentation Check
  - Relatively small change, but required updating a _lot_ of fixtures. the graphql commands could probably be handled differently to alleviate the need to update all of the fixtures. I want to look into making a fragment to store all of the conditional fragments for each check, such that adding new checks only means updating the single fragment.
  -


testdata/fixtures/check/update_has_documentation_request.json
```diff
@@ -0,0 +1,15 @@
...
+  "variables": {
+    "input": {
+      "id": "Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpIYXNPd25lci8yNDE4",
+      "name": "Hello World",
+      "enabled": true,
+      "categoryId": "Z2lkOi8vb3BzbGV2ZWwvQ2F0ZWdvcnkvNjA1",
+      "DocumentType": "api",
+      "DocumentSubtype": "openapi",
```
jasonopslevel, line 10: This isn't correct yet. The tests pass with these values in PascalCase vs camelCase - not sure why yet.

CONTRIBUTING.md
```diff
@@ -64,6 +64,16 @@ Next, to create a new change entry, in the root of the repository run: `changie
 
 Follow the prompts to create your change entry - remember this is what will show up in the changelog.  Changie registers the change in a .yaml file, and that file must be included in your pull request before we can release.
 
+### Auto Generating Types
+
+There is a code generator that helps maintain the various Enum types we use in graphql. It introspects the public Graphql Interface and keeps `enum.go` up to date. To use it:
+
+```
+go run gen.go
```
rocktavious, line 72: This idiomatic way is

```
go generate
```

If we ever add multiple generators this will ensure to run all of them

check.go
```diff
@@ -51,8 +52,8 @@ type CustomEventCheckFragment struct {
 }
 
 type ServiceOwnershipCheckFragment struct {
-	RequireContactMethod *bool                             `graphql:"requireContactMethod"`
-	ContactMethod        *ServiceOwnershipCheckContactType `graphql:"contactMethod"`
+	RequireContactMethod *bool        `graphql:"requireContactMethod"`
+	ContactMethod        *ContactType `graphql:"contactMethod"`
```
rocktavious, line 56: I recall this was put in for a reason - something about the field actually supporting more then just the ContactType because the API has it as a string.

Unless there is a reason to be undoing this lets not do this.

check.go
```diff
@@ -62,6 +63,11 @@ type HasRecentDeployCheckFragment struct {
 	Days int `graphql:"days"`
 }
 
+type HasDocumentationCheckFragment struct {
+	DocumentType    string `graphql:"documentType"`
```
rocktavious, line 67: These should be the Enum types so they stay strongly typed when someone on the golang side using this library goes to use them.

enum.go
```diff
@@ -130,28 +268,6 @@ func AllConnectiveEnum() []string {
 	}
 }
 
-// ContactType represents the method of contact.
-type ServiceOwnershipCheckContactType string
-
-const (
-	ServiceOwnershipCheckContactTypeAny         ServiceOwnershipCheckContactType = "any"          // Any contact method.
```
rocktavious, line 137: I think this was put in here because the service ownership check does not actually take an enum but there are only certain valid values.  And it also doesn't follow just the `ContactType` enum because "any" is not a valid `ContactType` but it is allowed in the `ServiceOwnershipCheck`

Seems like something that would need to be fixed on the Ruby side (using an Enum in that check type) before we can remove it here.

CC @maxim-opslevel to double check my understanding

CONTRIBUTING.md
```diff
@@ -64,6 +64,16 @@ Next, to create a new change entry, in the root of the repository run: `changie
...
+```
+
+Once the GraphQL API is made public, and before adding any other types - try the above command to pull down any types that can be auto-generated.
```
rocktavious, line 75: So the reason alot of Enums have been left out of the `enum.go` is because they are orphaned code.  There is alot still yet to be implemented and having the Enum's there would likely confuse an end user - "why is this enum here what is it used for?"

Now i'm not fully 100% against it I could be convinced otherwise - but i've been trying to keep the enum.go limited to only the things we actually end up using.  But I see how that can be hard with the generator.

So just relaying context - happy to allow in the unused enums if we feel strongly its a burden to maintain not committing unused ones.

testdata/fixtures/check/update_has_documentation_request.json
rocktavious: This has to do with the json marshaler - you'll need to add the json struct tags for the input types to have the marshaler used the cassing specific in the struct tag.

Then you'll beable to fix them here.

check.go
```diff
@@ -182,6 +188,20 @@ type CheckGitBranchProtectionUpdateInput struct {
 	CheckUpdateInput
 }
 
+type CheckHasDocumentationCreateInput struct {
+	CheckCreateInput
+
+	DocumentType    HasDocumentationTypeEnum    `graphql:"documentType"`
```
rocktavious, line 194: because these are "Input" they will be run through the JSON marshaler to go into the `variables` and that uses the `json` struct tags not the `graphql` struct tags so it needs to be

```
DocumentType    HasDocumentationTypeEnum    `json"documentType"`
```

Only the "return" types IE `HasDocumentationCheckFragment` needs the `graphql` struct tags because the graphql framework is marshaling the response from our API back into that struct and those tags tell it how to map the json back to the fields.

This is a very golang centric thing about data marshaling and structs - they are basically notes for the different marshalers to do their work and since we have to PascalCase the field names on the golang side so they are accessible by an enduser of this library we have to apply the struct tags to everything. :(

rocktavious (changes requested):
Thanks for the codebase wide `go fmt` and the contribution!  (I cannot wait until we fix the test fixtures - these massive search and replaces are getting out of hand)

Overall i'm a bit hesitant on the enum.go stuff - i mention why later but interested to hear your thoughts.

I've clarified some of the golang struct tags stuff you are running into so hopefully its more clear.

and then there are a few changes needed.

CONTRIBUTING.md
jasonopslevel: Ah ok, thank you for the context! I think it's confusing to have a generated file that we can only partially commit to. The file says "do not modify", but we still need to make manual modifications. Perhaps we could add the logic to omit certain fields in the generator that don't work (like the ContactType issue). 

I think it's fine having deprecated types in here though. I think the most important thing is the query interfaces, and the types just add validation on inputs/outputs so I would think most folks would just not interact with them.

check.go
jasonopslevel: Ah ok thanks again for this context! I was having a hard time figuring out why this was happening.

CONTRIBUTING.md
jasonopslevel: An example of where this confusion has already bit us is with the recent addition of `CheckServiceDependency` - It should have been `CheckTypeServiceDependency`. I will revert all of the other enum.go changes for now, but I _do_ think that one should go in imo. It's a simple rename and would prevent that change from happening every type we run `go generate`.

jasonopslevel:
Went through the comments and:

- fixed contributing.md: `go run generate.go` -> `go generate`
- fixed the grapqhl types, the fragments now properly use the Enum types instead of string, and the enums us the `json` tag so that they are marshalled correctly.
- removed the majority of the enum diff. I propsed to leave the `CheckServiceDependency` -> `CheckTypeServiceDependency` as I believe it is still correct (and it is only a simple small scope rename).

I believe `go generate` _should_ be fixed such that we can idempotently call it to help us generate the types, but that may be a decent amount of work. I may take a quick crack at reworking the Check tests to not need to do the find + replace, but in a future PR.

rocktavious (approved):
LGTM

Ya - I agree we should just beable to run `go generate` and it should be fine to commit I was just unsure if the unused enums would be bad.  I would hate to add exclusion logic to generation because then if someone actually needed that Enum it wouldn't generate and maybe they wouldn't know why.

Lets do the full enum.go update in a separate PR.

Happy to entertain a test fixture rewrite too - @eochoa260 and I talked about some options using go text templates but happy to see if you have any ideas too.

PR #76, opened 2022-08-10
Add alert source

terryopsleveltest:
Add `Get` functions and tests for alert source by both ID and external identifier

### DIFF
--- a/alert_source.go
+++ b/alert_source.go
@@ -5,19 +5,18 @@ import (
 )
 
 type AlertSourceExternalIdentifier struct {
-	Type 			 AlertSourceTypeEnum	`graphql:"type" json:"type"`
-	ExternalId string 							`graphql:"externalId" json:"externalId"`
+	Type       AlertSourceTypeEnum  `json:"type"`
+	ExternalId string               `json:"externalId"`
 }
 
 type AlertSource struct {
-	Name           string
-	Description    string
-	Id             graphql.ID
-	Type           AlertSourceTypeEnum
-	ExternalId     string
-	Integration    Integration
-	Url            string
-	Metadata       string
+	Name           string                 `graphql:"name"`
+	Description    string                 `graphql:"description"`
+	Id             graphql.ID             `graphql:"id"`
+	Type           AlertSourceTypeEnum    `graphql:"type"`
+	ExternalId     string                 `graphql:"externalId"`
+	Integration    Integration            `graphql:"integration"`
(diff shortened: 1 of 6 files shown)


alert_source.go
```diff
@@ -0,0 +1,57 @@
...
+
+type AlertSource struct {
+	Name           string
+	Description    string
+	Id             graphql.ID
+	Type           AlertSourceTypeEnum
+	ExternalId     string
+	Integration    Integration
+	Url            string
+	Metadata       string
```
rocktavious, line 20: Is there a size limit on metadata?  Looking at the metadata returned in our circular account it seems like this is probably not going to be generally useful to the CLI and Terraform.  So unless there is a good reason i'd recommend removing this field as it could potentially be alot of data to pull back from the API - especially once we implement the `List` function.

alert_source.go
```diff
@@ -0,0 +1,57 @@
+package opslevel
+
+import (
+	"github.com/shurcooL/graphql"
+)
+
+type AlertSourceExternalIdentifier struct {
+	Type 			 AlertSourceTypeEnum	`graphql:"type" json:"type"`
```
rocktavious, line 8: These technically only need the `json` struct tag as they are used for input only.  More context can be found here - https://github.com/OpsLevel/opslevel-go/pull/75#discussion_r939462528

alert_source.go
```diff
@@ -0,0 +1,57 @@
...
+	ExternalId string 							`graphql:"externalId" json:"externalId"`
+}
+
+type AlertSource struct {
+	Name           string
```
rocktavious, line 13: because this struct is used for unmarshaling the response data - for safety incase we need to change the field name - its best to put the `graphql` struct tags on these ones with the exact caseing used in our API responses.  Looking back I didn't do this with the User example I gave you and that was my bad - but it really should be there for safety (even if it works correctly now because it downcases the first letter).

rocktavious (changes requested):
Overall the PR works and does what needs to do.  I have one concern with the `metadata` field that i'm unsure if we should include and then some nits on the struct tags (up to you if you want to fix them).

The reason this is still "Request Changes" - is we need a changelog entry - you can read about adding that here - https://github.com/OpsLevel/opslevel-go/blob/main/CONTRIBUTING.md#changie-change-log-generation

rocktavious (approved):
LGTM! 🎉

PR #78, opened 2022-08-19
Add MaturityReport field to service

eochoa260:


maturity.go
```diff
@@ -25,6 +27,21 @@ func (s *MaturityReport) Get(category string) *Level {
 	return nil
 }
 
+func (c *Client) GetServiceMaturityWithAlias(alias string) (*ServiceMaturity, error) {
```
rocktavious, line 30: can we support ID or Alias with this call?  Using that NewIdentifier stuff that detects a GID from an Alias?

If not thats fine

rocktavious (approved):
1 question - but other then that LGTM

maturity.go
```diff
@@ -25,6 +27,21 @@ func (s *MaturityReport) Get(category string) *Level {
...
+	var q struct {
+		Account struct {
+			Service ServiceMaturity `graphql:"service(alias:$service)"`
+		}
+	}
+	v := PayloadVariables{
+		"service": graphql.String(alias),
+	}
+	if err := c.Query(&q, v); err != nil {
+		return nil, err
```
eochoa260, line 40: Is it possible to develop a failure case for this?  Would we have to find a way to get the client to return an error? Not sure of the best way to do this but tried looking into it for a bit.

rocktavious: ya we'd need an alternative test client that returns an http error rather then marshaling a fixture.  And then in that test you validate the error is bubbled up.  Its alot of effort for little return on test coverage.  But if you want to do it up to you.

I'd prefer we spend cycles on fixing our test fixtures first.  Once we have a patten it should be easy to go back and add these tests for error handling later - but that pattern might tie into our new test fixture pattern.

maturity.go
eochoa260: Let me try that.  I think that should work.

eochoa260: Is there something that has to be implemented on the API side to get this to work?  I tried passing in an IdentifierInput param on the GQL side and it's saying that it's not an accepted type.  I don't remember where else we did this for a query, only seeing mutation examples.

rocktavious: Looking at the graphql API docs for this it doesn't support the `IdentifierInput` yet only `id` or `alias` because each is nullable.  Here is the signature of the query:

```
service(id: ID alias: String): Service
```

Which means you'd need to pass null for `id` or `alias` depending on which value is given by the enduser using `GetServiceMaturityWithAlias`

Something like
```
	var q struct {
		Account struct {
			Service ServiceMaturity `graphql:"service(id: $id alias:$alias)"`
		}
	}
	v := PayloadVariables{
            "id": nil,
            "alias": nil,
        }
	if IsID(input) {
	    v["id"] = graphql.ID(input)
	} else {
	    v["id"] = graphql.String(input)
	}
	
```

PR #79, opened 2022-08-24
expand the definition of a runner job to include files

rocktavious:
This adds a new section the JobRun so that files can be injected to the container at runtime.

This changes the mutation to look like
```graphql
mutation ($id: ID!, $token: ID) {
  runnerGetPendingJob(runnerId: $id, lastUpdateToken: $token) {
    runnerJob {
      commands
      id
      image
      outcome
      status
      variables {
        key
        sensitive
        value
      }
      files {
        name
        contents
      }
    }
    lastUpdateToken
    errors {
      message
      path
    }
  }
}
```

here is an example yaml of a JobRun definition that would marshal into this.
```yaml
image: alpine/curl
commands:
  - /opslevel/check.sh
variables:
  - key: Secret
    value: "World!"
    sensitive: true
files:
  - name: check.sh
    contents: |
      #! /usr/bin/env sh
      echo "Hello World"
      sleep 5
      echo ${Secret}
```

### DIFF
--- a/job.go
+++ b/job.go
@@ -80,6 +80,11 @@ type RunnerJobVariable struct {
 	Value     string `json:"value"`
 }
 
+type RunnerJobFile struct {
+	Name     string `json:"name"`
+	Contents string `json:"contents"`
+}
+
 type RunnerJob struct {
 	Commands  []string             `json:"commands"`
 	Id        graphql.ID           `json:"id"`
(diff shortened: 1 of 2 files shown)


sabeerzaman (approved):
LGTM - changes here make sense, and successful test runs on staging 👍

PR #80, opened 2022-08-28
enable fixtures to be templated

rocktavious:
I've added go text templates with sprig to our testing library so we can reduce fixture content duplication. 🎉 

From here we can start to add more "test client" constructors so we can pass inline templates without having to load a fixture from disk.

We have a new directory `./testdata/templates/*` that contains the reuseable templates which get loaded up by autopilot and are then useable in our fixtures like this

```
{
  "data": {
    "serviceCreate": {
      "service": {{ template "service.tpl" }},
      "errors": []
    }
  }
}
```

All fixture files are loaded as if they are go text templates so if they contain templating they will get auto processed.  Also  any of the `./testdata/templates/*` files can have templating stanzas in them as well they will get recursively processed (thanks golang)

This means we can have shared values that are used across all our fixture files - see `./testdata/templates/common.tpl`


matthewbrahms (approved):
I _think_ I understand the mechanics of this, but looks good from what I can tell. 👍

eochoa260 (approved):

PR #82, opened 2022-09-15
Fixes API document update calls

rocktavious:
So the main impetunus for this change is the bugs found when testing - https://github.com/OpsLevel/terraform-provider-opslevel/pull/34

The main thing was the original code was implemented to an incorrect spec in GraphQL
what as implemented was non-nullable inputs for `docPath` and `docSource`
```
mutation($docPath:String!$docSource:ApiDocumentSourceEnum!$service:IdentifierInput!)
```
but the correct is with nullable inputs
```
mutation($docPath:String$docSource:ApiDocumentSourceEnum$service:IdentifierInput!)
```

### DIFF
--- a/clientGQL_test.go
+++ b/clientGQL_test.go
@@ -31,6 +31,11 @@ type GraphqlQuery struct {
 	Variables map[string]interface{} `json:",omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


eochoa260 (approved):
LGTM

PR #83, opened 2022-09-16
bugfix level update input

rocktavious:
This fixes 2 bugs with LevelUpdateInput mutation:
- The update was never able to work without setting Name
- The update was never able to set the description to and empty string

This will then lead to a change in the Terraform provider (and likely the CLI)


eochoa260 (approved):
LGTM

PR #84, opened 2022-09-21
fix: check err for nil

rene00:
client.Validate() returns err not a bool


rene00:
Fixes https://github.com/OpsLevel/opslevel-go/issues/85

rocktavious (approved):

PR #90, opened 2022-11-02
Update service ownership mutations with tag fields and allow creation of tags with team type

rauhut:
Adds fields required to support the create and update mutations for the service ownership check to include tag key and tag predicate in addition for supporting creating team tags


rocktavious (changes requested):
2 typo fixes need.

The one about string pointers i'm ok with not fixing since it requires a change for the bug you raised in slack.

rocktavious (approved):

PR #91, opened 2022-11-08
GraphQL having issues in mutations while updating checks

matthewbrahms:


rocktavious:
@matthewbrahms - looks like you were not based on the latest main before you committed so now there are a ton of weird diff entry lines.

PR #94, opened 2022-12-08
Adding repo grep check type

eochoa260:


testdata/fixtures/check/update_repo_grep_request.json
```diff
@@ -0,0 +1,16 @@
...
+    "variables": {
+      "input": {
+        "id": "Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpIYXNPd25lci8yNDE4",
+        {{ template "check_base_vars" }},
+        "directorySearch": true,
```
rocktavious, line 7: can you write a test that has this field missing and ensure we can generate that kind of request with our library.

testdata/fixtures/check/update_repo_grep_request.json
```diff
@@ -0,0 +1,16 @@
...
+        "filePaths": [
+          "**/go.mod"
+        ],
+        "fileContentsPredicate": {
```
rocktavious, line 11: same with this field - can we write a test thats doing an update where this field is missing and ensure we're able to generate that request with our library.

check_test.go
```diff
@@ -284,8 +312,8 @@ var testcases = map[string]struct {
 				CheckUpdateInput:     checkUpdateInput,
 				RequireContactMethod: ol.Bool(true),
 				ContactMethod:        &emailType,
-				TeamTagKey:           "updated_at",       
-				TeamTagPredicate:     &ol.PredicateUpdateInput{
+				TeamTagKey:           "updated_at",
+				TeamTagPredicate: &ol.PredicateUpdateInput{
```
rocktavious, line 316: why are these changes happening?

check_test.go
```diff
@@ -269,7 +297,7 @@ var testcases = map[string]struct {
 				RequireContactMethod: ol.Bool(true),
 				ContactMethod:        &slackType,
 				TeamTagKey:           "updated_at",
-				TeamTagPredicate:     &ol.PredicateInput{
+				TeamTagPredicate: &ol.PredicateInput{
```
rocktavious, line 300: this feels like an incorrect change - the spacing was removed which is against the go style

rocktavious (changes requested):
Overall - this looks good but can we harden the "nullability" stuff with actual tests that prove we can generate update calls with the nullable fields missing?

check_test.go
eochoa260: I'm not sure... goland is doing that automatically when I save it.  Tried moving it back but it keeps removing the whitespace.  Is it because it's setting it to a struct that has fields that are nested/indented on the next line so it just keeps that whole top line together? I haven't customized anything of the go styling so not really sure why it would act differently

check_test.go
eochoa260: Same as above here

check_test.go
rocktavious: ok lets ignore for now

testdata/fixtures/check/create_repo_grep_missing_directory_search_request.json
```diff
@@ -0,0 +1,19 @@
...
+    "variables": {
+      "input": {
+        "name": "Hello World",
+        "enabled": true,
+        "categoryId": "Z2lkOi8vb3BzbGV2ZWwvQ2F0ZWdvcnkvNjA1",
+        "levelId": "Z2lkOi8vb3BzbGV2ZWwvTGV2ZWwvMzE3",
+        "notes": "Hello World Check",
+        "directorySearch": false,
```
eochoa260, line 10: Not sure if this is the right thing to do but the graphql lib was putting this in by default even after I removed the field in the test case.

check_test.go
```diff
@@ -186,6 +186,57 @@ var testcases = map[string]struct {
...
+		},
+	},
+	"UpdateRepositoryGrepMissingFileContentsPredicate": {
+		fixture: "check/update_repo_grep_missing_file_contents_predicate",
+		body: func(c *ol.Client) (*ol.Check, error) {
+			return c.UpdateCheckRepositoryGrep(ol.CheckRepositoryGrepUpdateInput{
+				CheckUpdateInput: checkUpdateInput,
+				DirectorySearch:  ol.Bool(true),
+				Filepaths:        []string{"**/go.mod"},
+			})
```
rocktavious, line 237: I think what we actually need to test here is being able to "unset" `FileContentsPredicate` when its already set.  This is the usecase Terraform will do when its removed from the config and it needs to be turn from an existing predicate to "null" - basically "unset" a predicate.  I'm not confident our code as written can do that.

If this doesn't make sense lets sync on this Wednesday

eochoa260: Yeah, you're right, I see that FileContentsPredicate is unchanged when I run the mutation.  I'll take a look at it today to see if I can make the necessary changes.  I followed the pattern of other check types so sounds like we need to make this change in a lot of places.

rocktavious (approved):
LGTM - Ship it!

PR #95, opened 2022-12-22
1st pass at custom actions API’s

rocktavious:
Still missing Update and Delete - I'll try to get these done before the EOW


actions.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return &m.Payload.WebhookAction, FormatErrors(m.Payload.Errors)
+}
+
+// TODO: Not implemented in the API yet
+//func (client *Client) GetCustomAction(input IdentifierInput) (*CustomActionsTriggerDefinition, error) {
+
+//}
+
+func (client *Client) ListCustomActions(variables *PayloadVariables) (CustomActionsExternalActionsConnection, error) {
```
rocktavious, line 135: @eochoa260 - I took a slightly new approach here to pagination that uses a recursive function.  I kind of like this approach the best so far.

actions.go
```diff
@@ -0,0 +1,221 @@
...
+	if err := client.Query(&q, nil); err != nil {
+		return CustomActionsExternalActionsConnection{}, err
+	}
+	for q.Account.Actions.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Actions.PageInfo.End
+		resp, err := client.ListCustomActions(variables)
+		if err != nil {
+			return CustomActionsExternalActionsConnection{}, err
+		}
+		q.Account.Actions.Nodes = append(q.Account.Actions.Nodes, resp.Nodes...)
```
rocktavious, line 156: i'm not too thrilled about this section but i'm not sure of a better way to do it

actions_test.go
```diff
@@ -0,0 +1,160 @@
...
+	marshalled := ol.CustomActionsToHeaders(unstructured)
+	unmarshalled := ol.CustomActionsFromHeaders(structured)
+	// Assert
+	autopilot.Equals(t, structured, marshalled)
+	autopilot.Equals(t, unstructured, unmarshalled)
+}
+
+func TestCreateWebhookAction(t *testing.T) {
+	//Arrange
+	request := `{"query":
```
rocktavious, line 29: I created a new approach to test data that should hopefully reduct the amount of files we have to generate

I also then tried to use the templates in the "right" places to reduce the duplication

actions_test.go
```diff
@@ -0,0 +1,160 @@
...
+          ],
+          {{ template "no_pagination_response" }},
+          "totalCount": 2
+        }
+    }}}`
+
+	// An easy way to see the results of templating is by uncommenting this
+	//fmt.Print(Templated(request))
+	//fmt.Print(Templated(response))
+	//panic(1)
```
rocktavious, line 74: I left this in here as an example incase you need to debug the templated data that gets generated for the request and response

actions_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// Assert
+	autopilot.Equals(t, structured, marshalled)
+	autopilot.Equals(t, unstructured, unmarshalled)
+}
+
+func TestCreateWebhookAction(t *testing.T) {
+	//Arrange
+	request := `{"query": 
+		"mutation($input:CustomActionsWebhookActionCreateInput!){customActionsWebhookActionCreate(input: $input){webhookAction{{ template "custom_actions_request" }},errors{message,path}}}",
+		"variables":{"input":{"headers":"Content-Type=\"application/json\"","httpMethod":"POST","liquidTemplate":"{\"token\": \"XXX\", \"ref\":\"main\", \"action\": \"rollback\"}","name":"Deploy Rollback","webhookUrl":"https://gitlab.com/api/v4/projects/1/trigger/pipeline"}}
```
rocktavious, line 31: while i'm still not 100% happy with this method for generating test request and responses its MUCH better and should reduce the number of fixture files and limit us to only having template files

actions.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
+	var m struct {
+		Payload struct {
+			WebhookAction CustomActionsExternalAction
+			Errors        []OpsLevelErrors
+		} `graphql:"customActionsWebhookActionCreate(input: $input)"`
+	}
+	marshalledHeaders := CustomActionsFromHeaders(input.Headers)
+	input.HeadersString = &marshalledHeaders
```
rocktavious, line 120: This is the biggest hack right now but I see no other way to do this properly since they choose to go with a string in the API instead of a more complex object

actions_test.go
```diff
@@ -0,0 +1,160 @@
+package opslevel_test
+
+import (
+	ol "github.com/opslevel/opslevel-go/v2023"
+	"github.com/rocktavious/autopilot/v2022"
+	"testing"
+)
+
+func TestWebhookActionHeadersMarshalling(t *testing.T) {
```
rocktavious, line 9: I explicitly made this test to ensure we are generating the headers information consistently.

Long term I think this is going to cause the biggest problem for the terraform implementation.  Do your best to try and only use the map[string]string on the Terraform side both for the config and for the state.

I'm pretty sure we'll beable to make that work better then storing the headers in string format as thats going to be a nightmare for Terraform.

filters.go
```diff
@@ -22,11 +22,15 @@ type PredicateUpdateInput struct {
 	Value string            `json:"value,omitempty"`
 }
 
+type FilterId struct {
```
rocktavious, line 25: I made this type so that custom actions trigger definitions can return a more simple set of data for the filter.  This should hopefully reduce the data returned and then its up to the end user to turn that into a full object using `client.GetFilter`

actions_test.go
rocktavious: One thing we might want to consider soon is switching the go template delims `{{` and `}}` to `[[` and `]]` so as to not visually conflict with the GraphQL & Json syntax

actions.go
rocktavious: @DougEdey - How bad would it be if we wanted to use a hashmap/dictionary/custom object for headers vs this string syntax?  I just fear a TON of problems are going to crop up because of it.

common.go
```diff
@@ -101,6 +92,13 @@ func Bool(v bool) *bool {
 	return p
 }
 
+func HandleErrors(err error, errs []OpsLevelErrors) error {
```
rocktavious, line 95: I made this change to reduce the code paths and increase test coverage

common.go
```diff
@@ -115,8 +113,12 @@ func FormatErrors(errs []OpsLevelErrors) error {
 	return fmt.Errorf(strings.Join(errstrings, "\n"))
 }
 
+// NewId use "" to set "null" for ID input fields that can be nullified
 func NewID(id string) *graphql.ID {
-	output := graphql.ID(id)
+	var output graphql.ID
```
rocktavious, line 118: I had to make this change so I could have update mutations set ID's to a json value of "null" so you can properly "unset" something in an Update mutation

common.go
```diff
@@ -79,18 +79,9 @@ func NewIdentifier(value string) *IdentifierInput {
 	}
 }
 
-func NewEmptyString() *graphql.String {
-	s := graphql.String("")
-	return &s
-}
-
 func NewString(value string) *graphql.String {
-	var output *graphql.String = nil
```
rocktavious, line 88: This refactor is a bit scary - but i called it out in the changelog as a possible BREAKING CHANGE.

We'll want to see all the places in Terraform where we use this (if any) and ensure they still work

json.go
```diff
@@ -0,0 +1,36 @@
+package opslevel
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// JSON is a specialized map[string]string to support proper graphql serialization
+type JSON map[string]string
```
rocktavious, line 9: This is what the HEADERS field uses and AFAICT its a new type we've never had in the public API before.

clientGQL_test.go
```diff
@@ -58,23 +55,42 @@ func LogRaw() autopilot.RequestValidation {
 	}
 }
 
-func QueryValidation(t *testing.T, exp string) autopilot.RequestValidation {
-	return func(r *http.Request) {
-		q := Parse(r)
-		autopilot.Equals(t, exp, q.Query)
+func Templated(input string) string {
```
rocktavious, line 58: These work well - so in the future i'll move them to autopilot

actions_test.go
```diff
@@ -0,0 +1,299 @@
...
+  }}}`
+	responseErr := `{"data": {"customActionsTriggerDefinitionDelete": {
+      "deletedAlias": null,
+      "deletedId": null,
+      "errors": [{{ template "error1" }}]
+  }}}`
+
+	client := ABetterTestClient(t, "custom_actions/delete_trigger", request, response)
+	clientErr := ABetterTestClient(t, "custom_actions/delete_trigger_err", request, responseErr)
+	clientErr2 := ABetterTestClient(t, "custom_actions/delete_trigger_err2", request, "")
```
rocktavious, line 284: I did some extra error testing here to hit code paths we've never hit before to increase test coverage - we should probably move these somewhere "common" so they are not buried here

actions.go
```diff
@@ -0,0 +1,247 @@
...
+type CustomActionsTriggerDefinitionUpdateInput struct {
+	Id          graphql.ID      `json:"id"`
+	Name        *graphql.String `json:"name,omitempty"`
+	Description *graphql.String `json:"description,omitempty"`
+	Owner       *graphql.ID     `json:"ownerId,omitempty"`
+	Action      *graphql.ID     `json:"actionId,omitempty"`
+	Filter      *graphql.ID     `json:"filterId,omitempty"`
+}
+
+func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
```
eochoa260, line 96: I remember we talked about it but I'm forgetting why there isn't a GetWebhookAction function?

actions.go
```diff
@@ -0,0 +1,247 @@
...
+	v := PayloadVariables{
+		"input": input,
+	}
+	err := client.Mutate(&m, v)
+	return &m.Payload.WebhookAction, HandleErrors(err, m.Payload.Errors)
+}
+
+func (client *Client) DeleteWebhookAction(input IdentifierInput) error {
+	var m struct {
+		Payload ResourceDeletePayload `graphql:"customActionsWebhookActionDelete(resource: $input)"`
```
eochoa260, line 158: Updated this to use resource for the input payload.  Not sure if we want to ask dev team to change this to input to be more consistent with other mutations.

actions.go
```diff
@@ -0,0 +1,247 @@
...
+	v := PayloadVariables{
+		"input": input,
+	}
+	err := client.Mutate(&m, v)
+	return &m.Payload.TriggerDefinition, HandleErrors(err, m.Payload.Errors)
+}
+
+func (client *Client) DeleteTriggerDefinition(input IdentifierInput) error {
+	var m struct {
+		Payload ResourceDeletePayload `graphql:"customActionsTriggerDefinitionDelete(resource: $input)"`
```
eochoa260, line 240: Same as above here

eochoa260 (approved):
LGTM

PR #99, opened 2023-01-14
Fix check test and add new ID scalar type that works right

rocktavious:


check.go
```diff
@@ -145,10 +145,10 @@ type CheckUpdateInput struct {
 	Name     string        `json:"name,omitempty"`
 	Enabled  *bool         `json:"enabled,omitempty"`
 	EnableOn *iso8601.Time `json:"enableOn,omitempty"`
-	Category interface{}   `json:"categoryId,omitempty"`
-	Level    interface{}   `json:"levelId,omitempty"`
-	Owner    interface{}   `json:"ownerId,omitempty"`
-	Filter   interface{}   `json:"filterId,omitempty"`
+	Category interface{}   `json:"categoryId"`
```
eochoa260, line 148: So we'd have to change everything to an interface type?

scalar_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"encoding/json"
+	"github.com/hasura/go-graphql-client"
+	"github.com/opslevel/opslevel-go/v2023"
+	"github.com/rocktavious/autopilot/v2022"
+	"testing"
+)
+
+type TestInput struct {
+	Key1 opslevel.ID  `json:"key1"`
```
eochoa260, line 12: Is there a reason these are `opslevel.ID` instead of `interface{}` like the update struct?

scalar_test.go
```diff
@@ -0,0 +1,84 @@
...
+		Key3: &id2,
+		Key4: &id2,
+	}
+	// Act
+	buf1, err1 := json.Marshal(one)
+	buf2, err2 := json.Marshal(two)
+	buf3, err3 := json.Marshal(three)
+	// Assert
+	autopilot.Ok(t, err1)
+	autopilot.Equals(t, `{"key1":"","key3":null}`, string(buf1))
```
eochoa260, line 41: This is awesome! 🎉

check.go
rocktavious: well no.  This was a change from your branch.  If we are good with this new scalar type we'd change everything to an `opslevel.ID`

scalar_test.go
rocktavious: The updated struct was a change from your branch that I was based on when i made this change.  Ideally if we are good with this we'll use `opslevel.ID` everywhere from here on out

eochoa260 (approved):

PR #101, opened 2023-01-18
Add a paginated test client

rocktavious:
Usage of this to come in the near future

### DIFF
--- a/clientGQL_test.go
+++ b/clientGQL_test.go
@@ -97,6 +97,27 @@ func ATestClient(t *testing.T, endpoint string) *ol.Client {
 		autopilot.GraphQLQueryFixtureValidation(t, fmt.Sprintf("%s_request.json", endpoint)))))
 }
 
+type TestRequest struct {
+	Request  string
+	Response string
+}
+
+func RegisterPaginatedEndpoint(t *testing.T, endpoint string, requests ...TestRequest) string {
+	url := fmt.Sprintf("/LOCAL_TESTING/%s", endpoint)
+	requestCount := 0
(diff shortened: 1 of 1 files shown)


clientGQL_test.go
```diff
@@ -97,6 +97,27 @@ func ATestClient(t *testing.T, endpoint string) *ol.Client {
 		autopilot.GraphQLQueryFixtureValidation(t, fmt.Sprintf("%s_request.json", endpoint)))))
 }
 
+type TestRequest struct {
```
rocktavious, line 100: This allows the test writer to craft pairs of request and response mocks

clientGQL_test.go
```diff
@@ -97,6 +97,27 @@ func ATestClient(t *testing.T, endpoint string) *ol.Client {
...
+	Response string
+}
+
+func RegisterPaginatedEndpoint(t *testing.T, endpoint string, requests ...TestRequest) string {
+	url := fmt.Sprintf("/LOCAL_TESTING/%s", endpoint)
+	requestCount := 0
+	autopilot.Mux.HandleFunc(url, func(w http.ResponseWriter, r *http.Request) {
+		GraphQLQueryTemplatedValidation(t, requests[requestCount].Request)(r)
+		TemplatedResponse(requests[requestCount].Response)(w)
+		requestCount += 1
```
rocktavious, line 111: This count pair'd with the variadic function param `requests` allow us to create different response for each subsequent request - in the case of pagination the 1st request will return page info has next page == true and the 2nd will return a response where it is false.

eochoa260 (approved):
LGTM

PR #103, opened 2023-01-19
Update structs to use new ID type

eochoa260:


eochoa260:
I think I ended up with some overlap from Kyle's PR:  https://github.com/OpsLevel/opslevel-go/pull/100/files#diff-e357d0ea0b8a510f7e28034296ab9070e5f697b392b499e1a4f486d060ac4469
Need to probably rebase off of that one.

actions.go
```diff
@@ -75,9 +75,9 @@ type CustomActionsWebhookActionUpdateInput struct {
 type CustomActionsTriggerDefinitionCreateInput struct {
 	Name        string          `json:"name"`
 	Description *graphql.String `json:"description,omitempty"`
-	Owner       graphql.ID      `json:"ownerId"`
-	Action      *graphql.ID     `json:"actionId,omitempty"`
-	Filter      *graphql.ID     `json:"filterId,omitempty"`
+	Owner       ID              `json:"ownerId"`
+	Action      ID              `json:"actionId"`
```
rocktavious, line 79: I think we still want `omitempty` for `Action` and `Filter` since the are optional on create

actions.go
```diff
@@ -87,12 +87,12 @@ type CustomActionsTriggerDefinitionCreateInput struct {
...
 
 type CustomActionsTriggerDefinitionUpdateInput struct {
-	Id                     graphql.ID                                      `json:"id"`
+	Id                     ID                                              `json:"id"`
 	Name                   *graphql.String                                 `json:"name,omitempty"`
 	Description            *graphql.String                                 `json:"description,omitempty"`
-	Owner                  interface{}                                     `json:"ownerId,omitempty"`
-	Action                 interface{}                                     `json:"actionId,omitempty"`
-	Filter                 interface{}                                     `json:"filterId,omitempty"`
+	Owner                  *ID                                             `json:"ownerId"`
```
rocktavious, line 93: I think we want `omitempty` still for these because they are optional

check.go
```diff
@@ -173,21 +173,21 @@ type CheckAlertSourceUsageUpdateInput struct {
...
-	ServiceSelector  string      `json:"serviceSelector,omitempty"`
-	SuccessCondition string      `json:"successCondition,omitempty"`
-	Message          string      `json:"resultMessage,omitempty"`
-	PassPending      *bool       `json:"passPending,omitempty"`
-	Integration      *graphql.ID `json:"integrationId,omitempty"`
+	ServiceSelector  string `json:"serviceSelector,omitempty"`
+	SuccessCondition string `json:"successCondition,omitempty"`
+	Message          string `json:"resultMessage,omitempty"`
+	PassPending      *bool  `json:"passPending,omitempty"`
+	Integration      *ID    `json:"integrationId"`
```
rocktavious, line 190: i don't feel like this is correct - we should inspect this one and make sure the tests are good.

tags.go
```diff
@@ -213,15 +213,15 @@ func (client *Client) GetTagsForServiceWithAlias(alias string) ([]Tag, error) {
 	if serviceErr != nil {
 		return nil, serviceErr
 	}
-	return client.GetTagsForService(service.Id)
+	return client.GetTagsForService(ID(service.Id))
```
rocktavious, line 216: why is the extra `ID()` needed here?

rocktavious (commented):
Overall i seems good - just a few concerns.

As for the conflicts - ya lets work out tomorrow a merge and rebase and get these in ASAP because those common.go ones are going to affect all the PRs

tags.go
eochoa260: It shouldn't be needed after I converted service.Id to ID type. Must have forgotten to remove it.

check.go
eochoa260: Ok, I'll look at the tests for this one.

actions.go
eochoa260: Yeah, I figured it would be ok if we passed null for these so I didn't know if we wanted it to look more like the update struct for consistency or if we wanted to just omit the fields.

eochoa260: Oh sorry, this is on the update struct... I thought we decided we wanted to go with option 3 from the scalar tests you wrote and work out the logic on the terraform side?

rocktavious: option 3 should be the least likely one - for the most part for referencing IDs of other resources i think `Create` will use number 2 and `Update` will use number 4.

1 is likely needed for ID's that are the resource identifier itself.  I cannot think of an usecase where number 3 is actually needed.

eochoa260: Ah gotcha, I must have mixed them up.  I'll make those changes now.

rocktavious (approved):

PR #104, opened 2023-01-19
refactoring ListChecks and tests

matthewbrahms:
also shipped some templating created here backwards to user_test.go so we keep them aligned.


rocktavious (approved):
LGTM

PR #108, opened 2023-01-25
fix custom event check to properly allow for unsetting fields

rocktavious:


check_test.go
```diff
@@ -130,11 +130,30 @@ var testcases = map[string]struct {
...
+				PassPending:      ol.Bool(false),
+			})
+		},
+	},
+	"UpdateCustomEventNoMessage": {
+		fixture: "check/update_custom_event_no_message",
+		body: func(c *ol.Client) (*ol.Check, error) {
+			selector := ".metadata.name"
+			success := ".metadata.name"
+			message := ""
```
rocktavious, line 151: this test ensures we can "unset" the message which wasn't possible before

eochoa260 (approved):

PR #113, opened 2023-02-06
First pass at adding operation names

rocktavious:
This is the first pass at adding operation names to our opslevel-go client library which will help with metrics in Mixpanel.

The machinery is very small because the graphql library is doing the heavy lifting so the biggest question is - are these operation names the thing we want to slice by?


maxim-opslevel (approved):

eochoa260 (approved):
LGTM

PR #136, opened 2023-02-17
WIP: Refactor service and repo pagination

eochoa260:


service.go
```diff
@@ -161,9 +228,10 @@ func (client *Client) CreateService(input ServiceCreateInput) (*Service, error)
 	if err := client.Mutate(&m, v, WithName("ServiceCreate")); err != nil {
 		return nil, err
 	}
-	if err := m.Payload.Service.Hydrate(client); err != nil {
-		return &m.Payload.Service, err
-	}
+	// TODO: Figure out why this is breaking TestCreateService
```
rocktavious, line 231: Because Hydrate actually makes additional API calls now and the test client is only setup with 1 kind of response rather then the 2-3 extra ones it needs for tags, tools and repos.

PR #141, opened 2023-02-18
increase test coverage on tags methods

rocktavious:
The tags code has been around since the beginning and is some of the worst - i spent some time refactoring it and deprecating things and then wrote tests for uncovered lines.


testdata/templates/common.tpl
```diff
@@ -7,9 +7,9 @@
 {{- define "email1" }}kyle@opslevel.com{{ end }}
 {{- define "email2" }}edgar@opslevel.com{{ end }}
 {{- define "email3" }}matthew@opslevel.com{{ end }}
-{{- define "id1" }}MTIzNDU2Nzg5MTIzNDU2Nzg5{{ end }}
-{{- define "id2" }}OTg3NjU0MzIxOTg3NjU0MzIx{{ end }}
-{{- define "id3" }}OTg3NjU0Mzg5MTIzNjU0MzIx{{ end }}
+{{- define "id1" }}Z2lkOi8vMTIzNDU2Nzg5OTg3NjU0MzIx{{ end }}
```
rocktavious, line 10: these needed to be changed so they were proper IDs that properly pass through the `IsID` function and get properly detected as IDs vs Alias

tags.go
```diff
@@ -55,35 +51,28 @@ type TagDeleteInput struct {
...
-	}
-	input.Tags = append(input.Tags, TagInput{
-		Key:   key,
-		Value: value,
-	})
-	return client.AssignTags(input)
+	return client.AssignTags(alias, map[string]string{key: value})
 }
 
+// Deprecated: Use AssignTagsFor instead
```
rocktavious, line 64: The functions around tags where VERY messy - so i DRY'd them up and deprecated what we could.  I'm still not 100% happy but its a step in the right direction for increasing test coverage

matthewbrahms (approved):
looks good to me.

tags.go
```diff
@@ -204,56 +177,44 @@ func (conn *TagConnection) Hydrate(service ID, client *Client) error {
...
 	}
-	if err := q.Account.Service.Tags.Hydrate(service, client); err != nil {
-		return q.Account.Service.Tags.Nodes, err
-	}
-	return q.Account.Service.Tags.Nodes, nil
+	t, err := s.GetTags(client, nil)
+	return t.Nodes, err
 }
 
+// Deprecated: use client.GetService(id).GetTags(client, nil).TotalCount instead
```
eochoa260, line 210: Do we have to make the GetTags call? For some reason I thought GetService would return tags too since it's supposed to run the hydrate function.

eochoa260 (approved):
LGTM

tags.go
rocktavious: DOH! - you are correct!  I will fix this

rocktavious:
Closing this and merging tags and tools together because of test dependencies

PR #143, opened 2023-02-21
Refactor service pagination

eochoa260:


service.go
```diff
@@ -97,19 +97,142 @@ func (s *Service) HasTool(category ToolCategory, name string, environment string
...
-		return err
+
+	if s.Tools.PageInfo.HasNextPage {
+		variables := &PayloadVariables{}
+		(*variables)["after"] = s.Tools.PageInfo.End
+		resp, err := s.GetTools(client, variables)
+		if err != nil {
+			return err
+		}
+		s.Tools = resp
```
rocktavious, line 116: the tags code doesn't have this.  I cannot remember why or which way was better/more correct

service.go
```diff
@@ -97,19 +97,142 @@ func (s *Service) HasTool(category ToolCategory, name string, environment string
...
-		return err
+
+	if s.Repositories.PageInfo.HasNextPage {
+		variables := &PayloadVariables{}
+		(*variables)["after"] = s.Repositories.PageInfo.End
+		resp, err := s.GetRepositories(client, variables)
+		if err != nil {
+			return err
+		}
+		s.Repositories = resp
```
rocktavious, line 126: same as the tools/tags thing

service.go
```diff
@@ -238,105 +361,171 @@ func (client *Client) GetServiceCount() (int, error) {
...
-			if err := (&item).Hydrate(client); err != nil {
-				return err
-			}
-			conn.Nodes = append(conn.Nodes, item)
+		(*variables)["after"] = q.Account.Services.PageInfo.End
+		resp, err := client.ListServices(variables)
+		if err != nil {
+			return ServiceConnection{}, err
 		}
+		q.Account.Services.Nodes = append(q.Account.Services.Nodes, resp.Nodes...)
```
rocktavious, line 384: I think we need to loop through each service - call hydrate on it and then stuff it into the array

service.go
```diff
@@ -238,105 +361,171 @@ func (client *Client) GetServiceCount() (int, error) {
...
+		return ServiceConnection{}, err
+	}
+
+	for q.Account.Services.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Services.PageInfo.End
+		resp, err := client.ListServicesWithFramework(framework, variables)
+		if err != nil {
+			return ServiceConnection{}, err
 		}
+		q.Account.Services.Nodes = append(q.Account.Services.Nodes, resp.Nodes...)
```
rocktavious, line 412: same recursive hydrate problem

testdata/templates/service/repos.tpl
```diff
@@ -0,0 +1,604 @@
+{{- define "service_repo_node_1" }}
```
rocktavious, line 1: It would be really nice to DRY up this data and remove the weird references to things not part of our organization - but this is a minor tweak and i'm fine shipping with out it.  Just a comment when creating future test data.  Keep it DRY, clean, and reuseable.

rocktavious (changes requested):
A few questions/changes and 1 comment.

service.go
eochoa260: Removing it was correct since the Get* functions should populate each field on service object.  Removing them now.

rocktavious (approved):

PR #149, opened 2023-02-24
Added tests for remaining ListServiceWith* functions

eochoa260:


matthewbrahms (approved):

service_test.go
```diff
@@ -289,6 +288,29 @@ func TestCreateService(t *testing.T) {
...
+	}`
+	response := `{"data": {"serviceUpdate": {
+     "service": {{ template "service_1" }},
+     "errors": []
+ }}}`
+
+	client := ABetterTestClient(t, "service/update", request, response)
+
+	// Act
+	_, err := client.UpdateService(ol.ServiceUpdateInput{
```
rocktavious, line 305: we should validate the response object has a correct value

rocktavious (changes requested):
Just 1 adjustment on UpdateService test to ensure the deserializer works

service_test.go
eochoa260: Oh yeah, missed that.  Just added one.

rocktavious (approved):
LGTM

PR #156, opened 2023-03-01
Remove four unused methods in tools code

matthewbrahms:


rocktavious:
We cannot remove these yet because they were just deprecated between last release and now.  We need to give this time to bake.

PR #164, opened 2023-03-14
adding ListSystems code and tests!

matthewbrahms:
going solo on this, but I think I got it all!


rocktavious (approved):
Another flawless MR! 🎉

PR #167, opened 2023-03-17
Add domain get, create, and update functions

eochoa260:


domain.go
```diff
@@ -21,10 +21,18 @@ type DomainConnection struct {
 	TotalCount int      `json:"totalCount"`
 }
 
-type DomainInput struct {
+type DomainCreateInput struct {
 	Name        string `json:"name,omitempty"`
```
rocktavious, line 25: this has to be non-nullable because the api docs say its `String!` so it shouldn't have the `omitempty`

Also they were suppose to follow @eapache paradigm where the Create and Update mutations use the same "type" because the identifier is pulled out to a separate argument.

We might want to push back since the API is not public yet and they built the wrong thing and this makes our API very inconsistent.

rocktavious (approved):
1 comment about how they built the wrong API - might be worth pushing back on.

domain.go
eochoa260: Is there an example of mutations that use the same input type?  I couldn't find anything in opslevel-go.

eochoa260: Also, from looking at this, aren't createinput and updateinput different since one allows name to be null and the other doesn't?  How does that work if they're the same type?

PR #169, opened 2023-03-17
Add AssignSystem, Tags, and ChildSystem functions

eochoa260:


domain.go
```diff
@@ -36,15 +38,92 @@ type DomainUpdateInput struct {
...
+			} `graphql:"domain(input: $domain)"`
+		}
+	}
+	if s.Id == "" {
+		return nil, fmt.Errorf("Unable to get Systems, invalid domain id: '%s'", s.Id)
+	}
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+	(*variables)["domain"] = IdentifierInput{
```
eochoa260, line 54: I wanted to use NewIdentifier here since DomainId is an Identifier but it wasn't setting the query variable types correctly.  I have a feeling I'm missing something but not sure what.

domain.go
```diff
@@ -36,15 +38,92 @@ type DomainUpdateInput struct {
...
+			} `graphql:"domain(input: $domain)"`
+		}
+	}
+	if s.Id == "" {
+		return nil, fmt.Errorf("Unable to get Tags, invalid domain id: '%s'", s.Id)
+	}
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+	(*variables)["domain"] = IdentifierInput{
```
eochoa260, line 88: Same here

domain.go
```diff
@@ -36,15 +38,92 @@ type DomainUpdateInput struct {
...
-	return nil
+	var m struct {
+		Payload struct {
+			Domain Domain
+			Errors []OpsLevelErrors
+		} `graphql:"domainChildAssign(domain:$domain, childSystems:$childSystems)"`
+	}
+	childSystems := []IdentifierInput{}
+	for _, system := range systems {
+		childSystems = append(childSystems, IdentifierInput{Id: ID(system)})
```
eochoa260, line 117: Same here

domain.go
```diff
@@ -36,15 +38,92 @@ type DomainUpdateInput struct {
...
+	}
+	v := PayloadVariables{
+		"domain": IdentifierInput{
```
eochoa260, line 120: And here

domain_test.go
```diff
@@ -38,21 +38,158 @@ func TestDomainCreate(t *testing.T) {
 	autopilot.Equals(t, "Z2lkOi8vb3BzbGV2ZWwvRW50aXR5T2JqZWN0LzMw", string(result.Id))
 }
 
+func TestChildSystems(t *testing.T) {
+	// Arrange
+	requests := []TestRequest{
+		{`{"query": "query DomainChildSystemsList($after:String!$domain:IdentifierInput!$first:Int!){account{domain(input: $domain){childSystems(after: $after, first: $first){nodes{id,aliases,name,description,htmlUrl,owner{... on Group{alias,id},... on Team{alias,id}},parent{id,aliases,name,description,htmlUrl,owner{... on Group{alias,id},... on Team{alias,id}}}},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor},totalCount}}}}",
```
eochoa260, line 44: If I use NewIdentifier in the function, it shows `$domain:IdentifierInput` here instead of non-nullable type which gives me an error in GraphiQL explorer

domain.go
```diff
@@ -36,15 +38,91 @@ type DomainUpdateInput struct {
...
-	return nil
+	var m struct {
+		Payload struct {
+			Domain Domain
+			Errors []OpsLevelErrors
+		} `graphql:"domainChildAssign(domain:$domain, childSystems:$childSystems)"`
+	}
+	childSystems := []IdentifierInput{}
+	for _, system := range systems {
+		systemIdentifier := NewIdentifier(system)
```
eochoa260, line 117: was mainly concerned here because systems is a list of identifiers.  Tried updating with this.

domain.go
eochoa260: Shouldn't matter though since `s` should always have an ID?

rocktavious: This could be 2 things:
1.) `NewIdentifier` returns `*IdentifierInput` and here you are not using the pointer type but the value type
2.) the "ID" you are using in testing doesn't actually base64 decode properly so it thinks its an alias in the `IsID` check here - https://github.com/OpsLevel/opslevel-go/blob/5f92bd8065dd278c7f525f9c6100d0531b68a9ad/scalar.go#L39

domain.go
```diff
@@ -36,15 +38,91 @@ type DomainUpdateInput struct {
...
+	}
+	for q.Account.Domain.ChildSystems.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Domain.ChildSystems.PageInfo.End
+		resp, err := s.ChildSystems(client, variables)
+		if err != nil {
+			return nil, err
+		}
+		q.Account.Domain.ChildSystems.Nodes = append(q.Account.Domain.ChildSystems.Nodes, resp.Nodes...)
+		q.Account.Domain.ChildSystems.PageInfo = resp.PageInfo
+		q.Account.Domain.ChildSystems.TotalCount += resp.TotalCount
```
rocktavious, line 69: after the snafu we found out last week - lets move to not using `TotalCount` at all and instead use this
```suggestion
```

This also means we need to remove total count from the query using `graphql:"-"` struct tag so we can calculate it ourselves.

It also then means right before we return we need to do this
```
q.Account.Domain.ChildSystems.TotalCount = len(q.Account.Domain.ChildSystems.Nodes)
return &q.Account.Domain.ChildSystems, nil
```

domain.go
```diff
@@ -36,15 +38,91 @@ type DomainUpdateInput struct {
...
+			Errors []OpsLevelErrors
+		} `graphql:"domainChildAssign(domain:$domain, childSystems:$childSystems)"`
+	}
+	childSystems := []IdentifierInput{}
+	for _, system := range systems {
+		systemIdentifier := NewIdentifier(system)
+		childSystems = append(childSystems, *systemIdentifier)
+	}
+	v := PayloadVariables{
+		"domain":       IdentifierInput{Id: s.Id},
```
rocktavious, line 121: we "should" beable to use `NewIdentifier` here

domain.go
rocktavious: Might be handy to wrap this up into a utility function that takes in a `[]string` and return `[]IdentifierInput`. I suspect we'll need to do this kind of thing in the future with Infra and Custom Objects

domain.go
eochoa260: Went around and fixed this everywhere.

domain.go
eochoa260: Hmm, just realized you were saying to do this for a list but I did it for an individual string.  Adding this helper function now.  I can delete the one that we don't need.

domain.go
eochoa260: Updated all cases in Domains

domain.go
eochoa260: Updated this to use the new helper function but not sure if it's needed.  Making a few more changes.

eochoa260: Ok, just dereferencing the pointer now and have helper function for list case only.

domain.go
eochoa260: Ok, yeah don't need a helper function for single value but created one for the list case and implemented here.

scalar.go
```diff
@@ -46,6 +46,14 @@ func NewIdentifier(value string) *IdentifierInput {
 	}
 }
 
+func NewIdentifierList(values []string) []IdentifierInput {
```
rocktavious, line 49: make sure to write a test for this in `scalar_test.go` so we can keep up code coverage.

domain_test.go
```diff
@@ -38,21 +38,158 @@ func TestDomainCreate(t *testing.T) {
 	autopilot.Equals(t, "Z2lkOi8vb3BzbGV2ZWwvRW50aXR5T2JqZWN0LzMw", string(result.Id))
 }
 
+func TestDomainGetSystems(t *testing.T) {
+	// Arrange
+	requests := []TestRequest{
+		{`{"query": "query DomainChildSystemsList($after:String!$domain:IdentifierInput!$first:Int!){account{domain(input: $domain){childSystems(after: $after, first: $first){nodes{id,aliases,name,description,htmlUrl,owner{... on Group{alias,id},... on Team{alias,id}},parent{id,aliases,name,description,htmlUrl,owner{... on Group{alias,id},... on Team{alias,id}}}},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor},totalCount}}}}",
```
rocktavious, line 44: `totalCount` should no longer be in these queries

domain_test.go
```diff
@@ -38,21 +38,158 @@ func TestDomainCreate(t *testing.T) {
...
+				"data": {
+					"account": {
+						"domain": {
+							"childSystems": {
+								"nodes": [
+									{{ template "system1_response" }},
+									{{ template "system2_response" }}
+								],
+								{{ template "pagination_initial_pageInfo_response" }},
+								"totalCount": 2
```
rocktavious, line 62: `totalCount` can get removed - we should no longer be requesting this data in the response

domain.go
```diff
@@ -36,15 +38,83 @@ type DomainUpdateInput struct {
...
-	return nil
+	var m struct {
+		Payload struct {
+			Domain Domain
+			Errors []OpsLevelErrors
+		} `graphql:"domainChildAssign(domain:$domain, childSystems:$childSystems)"`
+	}
+	v := PayloadVariables{
+		"domain":       *NewIdentifier(string(s.Id)),
+		"childSystems": NewIdentifierList(systems),
```
rocktavious, line 114: This is pedantic but something worth knowing.  In golang they are `Array` which is different from a `List` data structure for a number of reasons.  Array's are static in size you have to allocate more memory to get a bigger array - you define them with a static capacity and then when you "append" to them you are actually creating a whole new array in memory (with increased capacity) and replacing your copy of it - thats why the golang syntax for appending to an array is equals with a method call and you have to pass in a reference the original array - under the hood its making a new array thats bigger, adding all the existing objects and then adding the new object.

Where as `List` is seen as a dynamically "expandable" array which does not exist in golang (unless someone wrote one) but does in other languages.  Under the hood those datatypes provide an interface ontop of array to make them easier to work with.  The main difference being if you took a pointer to an array, then did an `append` on it that original pointer would not have the appended data.

So while I don't care about the name of the function that much - in golang it would be better to be called `NewIdentifierArray` because that means something different then `NewIdentifierList`

eochoa260: Cool, thanks for the explanation.  I didn't realize there was a difference and still used to calling everything a list from my python days.  d

matthewbrahms: Quick question about arrays vs slices.  Isn't an `array` always a fixed entity, but `slices` are what can be like an array but also grow and get allocated more storage under the hood?  I think `slice` is a term often used interchangeably with `array`, but they are different if I remember correctly and am reading the difference right?

https://go.dev/blog/slices-intro

scalar.go
eochoa260: Ok, not sure if I'm testing it right but wrote something for this.  Open to any feedback though since I wasn't sure of best way to test it.

domain_test.go
eochoa260: Went around and removed total count but had to remove it from the tag connection too which will break other tests.  Made the required changes to get service's GetTags test to pass but had to sum it and now I'm wondering why we don't have to sum it on the domain side? 🤷

rocktavious: It would have been nicer to have done this in a separate PR so its isolated, easier to review and easier to rollback.

If you have time and its not too much trouble extracting this a separate PR would be worth it.

I'm not sure what you mean by the other part you are wondering about - i'll have to dig in to see/understand.

domain.go
rocktavious: Ya slices have some attributes close to what `Lists` are like in other languages - While i understand they work differently in golang to array's its just a bit "under the hood" difference with the automatic memory allocation and management that you don't really notice day to day.  for all intents and purposes golang does not truely have "List" built in like other languages.  For example in python you would do
```python
var myList = ["one", "two", "three"]
myList.Append("four")
```
The append method is directly on the type signifiying that it modifies the type in place where in golang you have the `=` operator in the syntax signifying you are making and assigning new memory.

Slices are memory efficient because instead of copying all the memory of the array into a new array they are making a lightweight "pointer like" thing to a chunk of the original array - its slightly weird and not like things in other languages.  I'm pretty sure you also cannot easily tell the difference between a slice and array in golang - i'm not 100% sure if reflection will tell you its an array or slice - where as reflection in other languages will definitely tell you a `List` is not the same type as an `Array`

domain_test.go
rocktavious: Just the refactor of all the other things that have totalcount that are not domains and systems.  I understand you concern with the fixtures of dependent types might still have total count and thats fine - we can refactor those later.

The problem is we could potentially merge the other refactor into main - but this domains stuff needs to be held out.

Also problematic is cluttering a PR with multiple changes - each PR should be small targeted changes - in hindsight it would be even better if each Create, Update, Get was each their own PR as the functionality should all be isolated and none of the code dependent upon each other.

Also in hindsight - We could stand to break our tickets down more like this so we make tickets to PRs more cleanly.

eochoa260: Yeah, that's actually a good point.  Would have been nice to split up the ticket and could have been easier to divide the work that way.  I'll make sure to make the totalCount changes on the system side when I finish out that PR and I'll open a separate PR for the dependent types.

rocktavious (approved):

PR #175, opened 2023-03-22
Make better error messages

rocktavious:
The tests should be self-explanatory but hopefully this gives us better error messages that are less obtuse and then we can tackle the actually Ruby API side.

### DIFF
--- a/common.go
+++ b/common.go
@@ -68,16 +68,16 @@ func FormatErrors(errs []OpsLevelErrors) error {
 		return nil
 	}
 
-	var errstrings []string
-	errstrings = append(errstrings, "OpsLevel API Errors:")
+	var sb strings.Builder
+	sb.WriteString("OpsLevel API Errors:\n")
 	for _, err := range errs {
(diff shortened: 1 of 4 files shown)


common.go
```diff
@@ -71,10 +71,13 @@ func FormatErrors(errs []OpsLevelErrors) error {
 	var errstrings []string
 	errstrings = append(errstrings, "OpsLevel API Errors:")
 	for _, err := range errs {
-		errstrings = append(errstrings, fmt.Sprintf("\t* %s", string(err.Message)))
+		if len(err.Path) > 0 && err.Path[0] == "base" {
```
eapache-opslevel, line 74: I would check `len() == 1` here, just in case we end up with sub-elements nested under `base` that have semantic value at some point? But it probably doesn't matter.

eapache-opslevel (approved):

common.go
rocktavious: good to know

PR #180, opened 2023-04-10
Blockout code for service dependencies

rocktavious:


service.go
```diff
@@ -30,6 +30,9 @@ type Service struct {
 	Tier                       Tier                         `json:"tier,omitempty"`
 	Timestamps                 Timestamps                   `json:"timestamps"`
 	Tools                      *ToolConnection              `json:"tools,omitempty"`
+
+	Dependencies *ServiceDependenciesConnection `graphql:"-"`
+	Dependents   *ServiceDependentsConnection   `graphql:"-"`
```
rocktavious, line 35: The only part i'm not happy about is this part - plugging this into the Hydrate function blows up the scope.   Not sure we want to take the on - but if we don't it means this addition is a bit different then our other sub resources like tools and repos.

dependencies.go
```diff
@@ -0,0 +1,149 @@
...
+	Edges    []ServiceDependentsEdge `graphql:"edges"`
+	PageInfo PageInfo
+}
+
+type ServiceDependencyKey struct {
+	Parent IdentifierInput `json:"sourceIdentifier"`
+	Child  IdentifierInput `json:"destinationIdentifier"`
+}
+
+type ServiceDependencyCreateInput struct {
```
eochoa260, line 41: Is there not going to be an update function?

service.go
eochoa260: Yeah I agree, it’s blown up the scope by in the past so I think it makes sense to tackle it in its own ticket if we have the time. Should be ok to add later if needed right?

dependencies.go
rocktavious: Theres a discussion about this with Ownership team.  If they can fit in making one we can implement it.

service.go
rocktavious: Ya the only downside is that if customers go to use the field it will be nil untill they call `service.GetDependents(client, nil)` and `service.GetDependencies(client, nil)`

Which is subpar.  But if we just auto call it for the customer then we'll be querying a ton of data all the time - even for a simple mutation such as "update the API docs path".

PR #190, opened 2023-05-09
hotfix for service hydrate

rocktavious:


rocktavious:
It was tested and it correctly works when creating a service (this is when the null happens) and that Get and List still work as expected)

PR #196, opened 2023-05-15
Add support for Custom Actions Trigger Definitions entity type

DougEdey:
OpsLevel added supported for Actions on the [Self Service page](https://docs.opslevel.com/docs/getting-started-with-custom-actions#using-actions-on-the-self-service-page) and this support needs to be added to the OpsLevel go client.

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -38,6 +38,7 @@ type CustomActionsTriggerDefinition struct {
 	Timestamps             Timestamps                                      `graphql:"timestamps"`
 	AccessControl          CustomActionsTriggerDefinitionAccessControlEnum `graphql:"accessControl"`
 	ResponseTemplate       string                                          `graphql:"responseTemplate"`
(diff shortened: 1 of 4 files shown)


actions.go
```diff
@@ -100,6 +102,7 @@ type CustomActionsTriggerDefinitionUpdateInput struct {
 	Published              *bool                                           `json:"published,omitempty"`
 	AccessControl          CustomActionsTriggerDefinitionAccessControlEnum `json:"accessControl,omitempty"`
 	ResponseTemplate       *string                                         `json:"responseTemplate,omitempty"`
+	EntityType             CustomActionsEntityTypeEnum                     `json:"entityType,omitempty"`
```
rocktavious, line 105: we might need to add an test to ensure we can "reset" this back to the default value.  I cannot remember if this is a terraform problem or an opslevel-go problem.

rocktavious (approved):
Other than the 1 thing with update i'd like to ensure -  its a good PR.  Thanks for the help!

PR #211, opened 2023-06-19
checks return rawNotes for notes

alan-opslevel:
- `Notes` returns `rawNotes` instead of `notes` from graphql so that chars like "<" and ">" are returned properly instead of "&gt ;"


alan-opslevel:
still wrangling with tests for this PR

alan-opslevel:
tl;dr for all the updated tests, updates query to query for `notes: rawNotes` instead of `notes`

matthewbrahms (approved):

PR #216, opened 2023-07-12
fix struct tag typo, update test accordingly

davidbloss:


testdata/fixtures/team/create_request.json
```diff
@@ -11,7 +11,6 @@
           "contacts":[
               {
                   "type":"slack",
-                  "displayName":"",
```
rocktavious, line 14: i think doing this request to our API would actually result in a display name of "".  We should ensure this change is not having us lose the ability have an empty string display name as well as during `update` we can wipe out the current display name.

rocktavious (changes requested):
I'd like to request more functional testing on this to ensure we don't loose functionality (and that we actually support all possible functionality of our API properly)

Also - we need a changie entry for this change - https://github.com/OpsLevel/opslevel-go/blob/main/CONTRIBUTING.md#changie-change-log-generation

rocktavious (approved):
LGTM

PR #217, opened 2023-07-13
Fix EntityOwner

rocktavious:
EntityOwner was not implemented correctly in our library and this papers around the fact that its non-easy to fix our API so opslevel-go has to do weird things.

### DIFF
--- a/owner.go
+++ b/owner.go
@@ -1,8 +1,26 @@
 package opslevel
 
+type EntityOwnerGroup struct {
+	Alias string `json:"alias,omitempty" graphql:"groupAlias:alias"`
+	Id    ID     `json:"id"`
+}
+
+type EntityOwnerTeam struct {
+	Alias string `json:"alias,omitempty" graphql:"teamAlias:alias"`
+	Id    ID     `json:"id"`
+}
+
 type EntityOwner struct {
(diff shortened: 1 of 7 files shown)


owner.go
```diff
@@ -1,8 +1,26 @@
...
+type EntityOwnerTeam struct {
+	Alias string `json:"alias,omitempty" graphql:"teamAlias:alias"`
+	Id    ID     `json:"id"`
+}
+
 type EntityOwner struct {
-	OnGroup GroupId `graphql:"... on Group"`
-	OnTeam  TeamId  `graphql:"... on Team"`
+	OnGroup EntityOwnerGroup `graphql:"... on Group"`
+	OnTeam  EntityOwnerTeam  `graphql:"... on Team"`
```
davidbloss, line 15: nit: thoughts on renaming `OnTeam` and `OnGroup` to `TeamFragment` and `GroupFragment` ?

davidbloss (approved):
LGTM! 🎉

owner.go
rocktavious: Because `OnTeam` and `OnGroup` are capitalized it means they are exported fields meaning customers could have used them already and I didn't want to break that contract too.  Its bad enough they are exported - maybe when/if we roll that back we can fix that too?

PR #221, opened 2023-07-14
Fixes for the CLI

rocktavious:
These are the things i needed to fix after putting opslevel-go into use in the CLI.  I was able to perform all the CLI commands then:
```
opslevel list infra-schemas
opslevel get infra-schemas Database
opslevel create infra
opslevel list infra
opslevel get infra XXXXX
opslevel update infra
opsleve delete infra XXXX
```

### DIFF
--- a/infra.go
+++ b/infra.go
@@ -14,13 +14,13 @@ type InfrastructureResourceSchemaConnection struct {
 }
 
 type InfrastructureResource struct {
-	Id          string      `json:"id"`
-	Aliases     []string    `json:"aliases"`
-	Name        string      `json:"name"`
-	Type        string      `json:"type" graphql:"type @include(if: $all)"`
-	Owner       EntityOwner `json:"owner" graphql:"owner @include(if: $all)"`
-	OwnerLocked bool        `json:"ownerLocked" graphql:"ownerLocked @include(if: $all)"`
-	Data        JSON        `json:"data" scalar:"true" graphql:"data @include(if: $all)"`
+	Id          string         `json:"id"`
+	Aliases     []string       `json:"aliases"`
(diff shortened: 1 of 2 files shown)


infra.go
```diff
@@ -30,21 +30,21 @@ type InfrastructureResourceConnection struct {
...
 type InfrastructureResourceInput struct {
-	Type         *string                              `json:"providerResourceType,omitempty"`
+	Type         *string                              `json:"providerResourceType,omitempty" yaml:"providerResourceType"`
 	Schema       *InfrastructureResourceSchemaInput   `json:"schema,omitempty"`
-	ProviderData *InfrastructureResourceProviderInput `json:"providerData,omitempty"`
-	Owner        *ID                                  `json:"ownerId,omitempty"`
-	Data         JSON                                 `json:"data,omitempty" scalar:"true"`
+	ProviderData *InfrastructureResourceProviderInput `json:"providerData,omitempty" yaml:"providerData"`
+	Owner        *ID                                  `json:"ownerId,omitempty" yaml:"owner"`
+	Data         map[string]any                       `json:"data,omitempty" yaml:"data"`
```
davidbloss, line 47: I noticed `InfrastructureResourceInput`'s `Data` struct tags no longer include `scalar:"true"` but `InfrastructureResource`'s `Data` struct tags above do include `scalar:"true"`. I'm guessing this is fine since there aren't any `graphql` tags here?

davidbloss (approved):
Solid work! LGTM! 🎉

infra.go
rocktavious: Ya - just oversight - we'll be undoing this once the real fix is in

PR #222, opened 2023-07-15
infra create and update v2

rocktavious:
This change proposes a different signature for the create and update infra methods because the ones based on the API are not intuitive for customers when I put them into use and IMO this provides a much better UX when using the library.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -101,7 +101,7 @@ func (c *Cacher) TryGetInfrastructureSchema(alias string) (*InfrastructureResour
 }
 
 func (c *Cacher) doCacheTiers(client *Client) {
-	log.Info().Msg("Caching 'Tier' lookup table from API ...")
+	log.Debug().Msg("Caching 'Tier' lookup table from API ...")
 
 	data, dataErr := client.ListTiers()
 	if dataErr != nil {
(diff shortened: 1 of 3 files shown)


cache.go
```diff
@@ -101,7 +101,7 @@ func (c *Cacher) TryGetInfrastructureSchema(alias string) (*InfrastructureResour
 }
 
 func (c *Cacher) doCacheTiers(client *Client) {
-	log.Info().Msg("Caching 'Tier' lookup table from API ...")
+	log.Debug().Msg("Caching 'Tier' lookup table from API ...")
```
rocktavious, line 104: These info logs were screwing things up in the CLI

infra.go
```diff
@@ -56,7 +56,36 @@ type InfrastructureResourceInput struct {
 	Data         JSON                                 `json:"data,omitempty" yaml:"data" scalar:"true"`
 }
 
-func (client *Client) CreateInfrastructure(input InfrastructureResourceInput) (*InfrastructureResource, error) {
+type InfraProviderInput struct {
```
rocktavious, line 59: These new input types don't exist in our API but they provide a WAY better customer UX when using the library.  See the code in the CLI and Terraform as implementation examples.

infra.go
```diff
@@ -56,7 +56,36 @@ type InfrastructureResourceInput struct {
...
+
+type InfraInput struct {
+	Schema   string              `json:"schema" yaml:"schema"`
+	Owner    *ID                 `json:"owner" yaml:"owner"`
+	Provider *InfraProviderInput `json:"provider" yaml:"provider"`
+	Data     JSON                `json:"data" yaml:"data" scalar:"true"`
+}
+
+func (client *Client) CreateInfrastructure(input InfraInput) (*InfrastructureResource, error) {
+	i := InfrastructureResourceInput{
```
rocktavious, line 74: We do this translation layer in each function now.

davidbloss (approved):
I'm liking the cli UX improvements 🤩 
LGTM! 🎉

PR #226, opened 2023-07-21
Add nested filter support and change graphql client

taimoor-at-opslevel:
## Add nested filter support

Allow handling of filters that call other filters (i.e. "nested filters") using `and`/`or` conditions.

The list, delete, get operations should work the same way for nested filters, so no updates were made there.

Validated no changes were needed to filters.go, tested the new create/update queries live using graphiql.

## Changelog

1. Added test to create a nested filter (plus add template file for this)
1. Added test to update a nested filter (plus add template file for this)
1. Replaced https://github.com/shurcooL/graphql/ with https://github.com/hasura/go-graphql-client which added 1 dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
 	github.com/huandu/xstrings v1.3.1 // indirect
 	github.com/imdario/mergo v0.3.11 // indirect
-	github.com/klauspost/compress v1.15.14 // indirect
+	github.com/klauspost/compress v1.16.7 // indirect
 	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
 	github.com/mitchellh/copystructure v1.0.0 // indirect
--- a/.changes/unreleased/Feature-20230726-144153.yaml
+++ b/.changes/unreleased/Feature-20230726-144153.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: As a user, I can configure nested filters
+time: 2023-07-26T14:41:53.475416-04:00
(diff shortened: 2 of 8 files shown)


enum.go
```diff
@@ -271,9 +271,11 @@ const (
 )
 
 // All ConnectiveEnum as []string
-var AllConnectiveEnum = []string{
-	string(ConnectiveEnumAnd),
-	string(ConnectiveEnumOr),
+func AllConnectiveEnum() []string {
```
rocktavious, line 274: It looks like the generator didn't get updated to correctly create static global vars for the `all` lists (terraform uses these)

We need to undo the transformation to a function and fix the generator.

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/mitchellh/copystructure v1.0.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.0 // indirect
 	github.com/shopspring/decimal v1.2.0 // indirect
+	github.com/shurcooL/graphql v0.0.0-20230714182844-3e04114ae69a // indirect
```
rocktavious, line 32: looks like the generator still uses the outdated library - we should update that and then it won't need this in go.mod

taimoor-at-opslevel: Changed to https://github.com/hasura/go-graphql-client

enum.go
taimoor-at-opslevel: Rebased to `main` which has https://github.com/OpsLevel/opslevel-go/pull/227

.changes/v2023.7.25.md
```diff
@@ -0,0 +1,7 @@
+## [July 25, 2023](https://github.com/OpsLevel/opslevel-go/compare/v2023.7.17...v2023.7.25)
```
rocktavious, line 1: Ahh sorry this is my bad - you went too far with changie - when adding new features or making refactors - in the golang tools all you need to do is `changie new` the rest of the process is automated in our release.  I know it differs from self-hosted but thats because we couldn't/didn't fully automated that on purpose.

rocktavious (changes requested):
See my note about changie.

.changes/v2023.7.25.md
taimoor-at-opslevel: Fixed this

rocktavious (approved):

PR #233, opened 2023-08-14
Scorecard Support

taimoor-at-opslevel:
https://github.com/OpsLevel/team-platform/issues/31

CLI: https://github.com/OpsLevel/cli/pull/154


scorecards.go
```diff
@@ -0,0 +1,124 @@
...
+
+type ScorecardConnection struct {
+	Nodes      []Scorecard `graphql:"nodes"`
+	PageInfo   PageInfo    `graphql:"pageInfo"`
+	TotalCount int         `graphql:"totalCount"`
+}
+
+type ScorecardInput struct {
+	Name        string          `graphql:"name" json:"name"`
+	Description string          `graphql:"description" json:"description,omitempty"`
```
rocktavious, line 38: Just doing some early code review - on "Input" types when a field in the API is nullable we need to use pointers so the graphql framework will generate it correctly because you need 3 states - nil, empty string, filled string.  With just a string (value type) you only have 2 states - empty string, filled string.   

The `json` struct tag omitempty also works differently when its `string` vs `*string`
for `string` the omitempty will omit the field from the json when its the empty string state
for `*string` the omitempty will omit the field from the json when its the nil state but empty string state will still include the field in the json it will just be an empty string value which effectively works as an "unset the field" during an "update" mutation.

Its VERY complicated and it probably the hardest thing about making "Input" types.

scorecards.go
```diff
@@ -0,0 +1,124 @@
...
+	Checks         []string `graphql:"checks"`
+	Description    string   `graphql:"description"`
+	Filter         string   `graphql:"filter"`
+	Href           string   `graphql:"href"`
+	Levels         []string `graphql:"levels"`
+	Name           string   `graphql:"name"`
+	Owner          string   `graphql:"owner"`
+	PassingChecks  int      `graphql:"passingChecks"`
+	ServiceCount   int      `graphql:"serviceCount"`
+	ServicesReport string   `graphql:"servicesReport"`
```
rocktavious, line 25: we will not be implementing this field

scorecards.go
taimoor-at-opslevel: Thanks - I'll update this.

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+type ScorecardInput struct {
+	Name        string  `graphql:"name" json:"name"`
+	Description *string `graphql:"description" json:"description,omitempty"`
+	// TODO: API shouldn't use IdentifierInput for Owner if we support groups and teams
+	// if an alias is passed, is that for a Group or a Team?
+	// on Domain and System objects, they only support ID type as the owner, not InputIdentifier
+	// if this is changed, update the GetScorecard test
+	// TODO: a way to get linked checks from a scorecard
+	// requires a change in API and structs
+	Owner    IdentifierInput `graphql:"owner" json:"owner"`
```
rocktavious, line 45:
```suggestion
	Owner    ID `graphql:"owner" json:"ownerId"`
```

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+	TotalChecks   int         `graphql:"totalChecks"`
+}
+
+type ScorecardConnection struct {
+	Nodes      []Scorecard `graphql:"nodes"`
+	PageInfo   PageInfo    `graphql:"pageInfo"`
+	TotalCount int         `graphql:"totalCount"`
+}
+
+type ScorecardCheckConnection struct {
```
rocktavious, line 30: This type is unused

scorecards_test.go
```diff
@@ -0,0 +1,63 @@
...
+	sc, err := client.UpdateScorecard(ol.IdentifierInput{Id: "scorecard-id"}, ol.ScorecardInput{
+		Name:  "updated scorecard",
+		Owner: ol.IdentifierInput{Id: "team-id"},
+	})
+
+	autopilot.Ok(t, err)
+	autopilot.Equals(t, "updated scorecard", sc.Name)
+	autopilot.Equals(t, ol.ID("team-id"), sc.Owner.OnTeam.Id)
+}
+
```
rocktavious, line 39: Looks like we are missing a test for `ListScorecards`

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+type Scorecard struct {
+	ScorecardId
+
+	Description   string      `graphql:"description"`
+	Filter        Filter      `graphql:"filter"`
+	Name          string      `graphql:"name"`
+	Owner         EntityOwner `graphql:"owner"`
+	PassingChecks int         `graphql:"passingChecks"`
+	ServiceCount  int         `graphql:"serviceCount"`
+	TotalChecks   int         `graphql:"totalChecks"`
```
rocktavious, line 21: What do you think about renaming this to `ChecksCount` to match `ServiceCount` ?

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+		return ScorecardConnection{}, err
+	}
+	for q.Account.Scorecards.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Scorecards.PageInfo.End
+		resp, err := client.ListScorecards(variables)
+		if err != nil {
+			return ScorecardConnection{}, err
+		}
+		q.Account.Scorecards.Nodes = append(q.Account.Scorecards.Nodes, resp.Nodes...)
+		q.Account.Scorecards.PageInfo = resp.PageInfo
```
rocktavious, line 98: One thing that missing here if you look at other lists functions like Systems or Domains - is `TotalCount` which we pre-calculate (we don't use the API field on the connection type because its not always correct - don't ask why its a long story)

I'm fine if we don't want to add this till we have a usecase.  I've just been trying to keep parity between all the functions to make the future codegen adoption easier.

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	v := PayloadVariables{
+		"scorecard": scorecard,
+		"input":     input,
+	}
+	err := client.Mutate(&m, v, WithName("ScorecardUpdate"))
+	return &m.Payload.Scorecard, HandleErrors(err, m.Payload.Errors)
+}
+
+func (client *Client) DeleteScorecard(input IdentifierInput) (ID, error) {
```
rocktavious, line 118: Historically - i've been having these take in a `string` and then having the function internally do the transformation to `IdentifierInput` - this makes it easier in the CLI and Terraform if the signature is a string.

scorecards.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return ScorecardConnection{}, err
+		}
+		q.Account.Scorecards.Nodes = append(q.Account.Scorecards.Nodes, resp.Nodes...)
+		q.Account.Scorecards.PageInfo = resp.PageInfo
+	}
+	return q.Account.Scorecards, nil
+}
+
+func (client *Client) UpdateScorecard(scorecard IdentifierInput, input ScorecardInput) (*Scorecard, error) {
```
rocktavious, line 103: Same thing for the scorecard identifier here - if its a string we support both ID or Alias and the end user doesn't have to care - we handle it inside the function

scorecards.go
taimoor-at-opslevel: I'll add this change to scorecards, fwiw lots of other objects require `IdentifierInput` over `string`, like custom actions

scorecards.go
taimoor-at-opslevel: Looks like we can't `omitempty` here, since the field is required in `ScorecardInput`

scorecards_test.go
taimoor-at-opslevel: Added!

scorecards.go
taimoor-at-opslevel: I added this as a TODO item on line 88 since we need to get this out early this week

scorecards.go
rocktavious: Ok - then it should be a value type not a pointer type

testdata/templates/scorecards.tpl
```diff
@@ -0,0 +1,58 @@
...
+    "query": "query ScorecardGet($input:IdentifierInput!){account{scorecard(input: $input){aliases,id,description,filter{connective,htmlUrl,id,name,predicates{key,keyData,type,value}},name,owner{... on Group{groupAlias:alias,id},... on Team{teamAlias:alias,id}},passingChecks,serviceCount,totalChecks}}}",
+    "variables": {"input":{"id":"Z2lkOi8vMTIzNDU2Nzg5MTAK"}}
+}{{ end }}
+
+{{- define "scorecard_get_response" }}{
+    "data": {"account":{"scorecard":{"id":"Z2lkOi8vMTIzNDU2Nzg5MTAK", "name":"existing scorecard","owner":{"id":"Z2lkOi8vMTIzNDU2Nzg5Cg=="}}}}
+}{{ end }}
+
+{{- define "scorecard_list_request" }}{
+    "query": "query {account {scorecards {nodes {aliases,id,description,filter{connective,htmlUrl,id,name,predicates{key,keyData,type,value}},name,owner{... on Group{groupAlias:alias,id},... on Team{teamAlias:alias,id}},passingChecks,serviceCount,totalChecks}}}}",
```
rocktavious, line 38: this seems unused?

scorecards.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	for q.Account.Scorecards.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Scorecards.PageInfo.End
+		resp, err := client.ListScorecards(variables)
+		if err != nil {
+			return ScorecardConnection{}, err
+		}
+		q.Account.Scorecards.Nodes = append(q.Account.Scorecards.Nodes, resp.Nodes...)
+		q.Account.Scorecards.PageInfo = resp.PageInfo
+		// TODO: pre-calculate the TotalCount so that it's more correct, see domains.go/systems.go
```
rocktavious, line 88: Looks like a missing todo - ping @davidbloss on this one as I just ran him through it with Secrets too.

rocktavious (changes requested):
A couple more changes i see

scorecards.go
taimoor-at-opslevel: Good point, just changed it so that `ownerId` is no longer type `*ID`, it's now `ID`.

testdata/templates/scorecards.tpl
```diff
@@ -0,0 +1,58 @@
...
+    {{ template "pagination_second_query_variables" }}
+}{{ end }}
+
+{{- define "scorecard_1_response" }}
+    "id":"Z2lkOi8vMTIzNDU2Nzg5MTAK",
+    "name":"first scorecard",
+    "owner":{"id":"Z2lkOi8vMTIzNDU2Nzg5Cg=="}
```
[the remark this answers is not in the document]
taimoor-at-opslevel: I updated the tests for get/list which should cover all the top level fields. Tests for create/update test all the scorecard fields that are found in the ScorecardInput

scorecards.go
```diff
@@ -0,0 +1,122 @@
...
+type ScorecardId struct {
+	Aliases []string `graphql:"aliases"`
+	Id      ID       `graphql:"id"`
+}
+
+type Scorecard struct {
+	ScorecardId
+
+	Description   string      `graphql:"description"` // optional
+	Filter        Filter      `graphql:"filter"`      // optional
```
taimoor-at-opslevel, line 16: This should be a pointer with `omitempty` because it is causing CLI to show phantom filters when there aren't any.

```
    "Filter": {
        "Connective": "",
        "HtmlURL": "",
        "Id": null,
        "Name": "",
        "Predicates": null
    },
```

Can do the same for `Description`

rocktavious: Historically - i don't use pointers on non Input types so things are initialized.  For dependent resources like this usually we do a `.Id != nil` check if you want to know if the object is set or not.

The `omitempty` won't leave out the string pointers - its a JSON struct tag so its only used for marshaling a struct to JSON not the other way around.

I'm open to change this though if you think it doesn't make sense or would be better being fully nil if its not in the response payload.

rocktavious (approved):
While this is 👍 IMO i would not merge it yet because the API is not actually public.  I'll need to check with @reidan for when we should merge this.

scorecards.go
taimoor-at-opslevel: That makes sense to me. Changing this would be outside of the PR's scope, so let's leave it as-is.

PR #240, opened 2023-08-28
Fix client library to respect `Retry-After`header

rocktavious:
Investigation based on client side rate limiting.

This PR does 2 things:
- Make the handling of the API token more sane by passing it through the ClientSettings
- Remove the OAuth HTTP client wrapper and hardcoded timeout which is inadvertently messing with go-retryablehttp Backoff and Retry strategies.

This fully delegates the retry and backoff to the go-retryablehttp so they actually work properly and testing shows that when the server gives back the headers for `Retry-After` the client properly retries and errors stop happening.

This was tested using this script

```sh
go build -o ./cli
chmod +x  ./cli
  for i in `seq 1 30`; do
    for i in `seq 1 200`; do
      ./cli graphql -q='query Raw{ account { id } }' &
    done;
  sleep 5
  done;
```

Prior to this change we see errors in the output but after this change we don't see errors in the output ever.

### DIFF
--- a/client.go
+++ b/client.go
@@ -10,6 +10,7 @@ import (
 
 type ClientSettings struct {
 	url      string
+	token    string
 	timeout  time.Duration
 	retries  int
 	headers  map[string]string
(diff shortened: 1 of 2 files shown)


taimoor-at-opslevel (approved):

davidbloss (approved):
LGTM 🚀

PR #241, opened 2023-08-28
add linting to test CI workflow, add task setup

davidbloss:
Add `task setup` as a generic local development task.
- This setups up `gofumpt` and `golangci-lint` locally and in CI pipelines
- This also serves as a place to add new tools when introduced

`task lint` added to CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,6 @@ jobs:
           task setup
           task lint
           task test
-          grep -v "/enum.go" coverage.out > coverage.txt
       - name: Upload Coverage
         run: |-
           bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,10 +19,17 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: "1.21"
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Run Tests
         run: |-
-          go test -race -coverprofile=coverage.out -covermode=atomic -v ./...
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -19,10 +19,17 @@ jobs:
...
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Run Tests
         run: |-
-          go test -race -coverprofile=coverage.out -covermode=atomic -v ./...
-          cat coverage.out | grep -v "/enum.go" > coverage.txt
+          task setup
+          task lint
+          task test
+          grep -v "/enum.go" coverage.out > coverage.txt
```
rocktavious, line 32: any reason to no bake this into `task test` ?

davidbloss: Not really :)
I can add it in

Taskfile.yml
```diff
@@ -36,8 +25,43 @@ tasks:
...
+    cmds:
+      - task: install-go-tool
+        vars: { GO_TOOL: "gofumpt", GO_TOOL_PATH: "mvdan.cc/gofumpt@latest" }
+
+  install-golangci:
+    desc: go install "golangci-lint" and set GOBIN if not set
+    silent: true
+    cmds:
+      - task: install-go-tool
+        vars: { GO_TOOL: "golangci-lint", GO_TOOL_PATH: "github.com/golangci/golangci-lint/cmd/golangci-lint@v1.54.2" }
```
rocktavious, line 67: can we add a comment to why the version is pinned vs using `@latest` ?

davidbloss: I thought the golangci-lint page had a reason for pinning this, but I didn't see it. So I'm setting it to `@latest`

rocktavious: 🎉

rocktavious (approved):

PR #245, opened 2023-09-11
Db/add extended team access

davidbloss:
WIP: Add extendedTeamAccess to `CustomActionsTriggerDefinition` with tests

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -41,6 +41,39 @@ type CustomActionsTriggerDefinition struct {
 	EntityType             CustomActionsEntityTypeEnum                     `graphql:"entityType"`
 }
 
+func (c *CustomActionsTriggerDefinition) ExtendedTeamAccess(client *Client, variables *PayloadVariables) (*TeamConnection, error) {
+	var q struct {
+		Account struct {
+			CustomActionsTriggerDefinition struct {
+				ExtendedTeamAccess TeamConnection `graphql:"extendedTeamAccess(after: $after, first: $first)"`
+			} `graphql:"customActionsTriggerDefinition(input: $input)"`
+		}
+	}
+	if c.Id == "" {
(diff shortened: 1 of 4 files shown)


actions_test.go
```diff
@@ -383,3 +384,35 @@ func TestDeleteTriggerDefinition(t *testing.T) {
...
+	}
+
+	client := APaginatedTestClient(t, "custom_actions/list_extended_team_access", requests...)
+	id1 := *ol.NewID("Z2lkOi8vMTIzNDU2Nzg5OTg3NjU0MzIx")
+	trigger := ol.CustomActionsTriggerDefinition{Id: id1}
+
+	// Act
+	resp, err := trigger.ExtendedTeamAccess(client, nil)
+	if err != nil {
+		panic(err)
```
rocktavious, line 410: this should be removed

actions.go
```diff
@@ -103,6 +137,7 @@ type CustomActionsTriggerDefinitionUpdateInput struct {
 	AccessControl          CustomActionsTriggerDefinitionAccessControlEnum `json:"accessControl,omitempty"`
 	ResponseTemplate       *string                                         `json:"responseTemplate,omitempty"`
 	EntityType             CustomActionsEntityTypeEnum                     `json:"entityType,omitempty"`
+	ExtendedTeamAccess     *[]IdentifierInput                              `json:"extendedTeamAccess,omitempty"`
```
rocktavious, line 140: This was the field I wanted extra tests on to ensure we can do the 3 states and that we generate the input variables correctly:
- not present doesn't show up in the variables
- empty array unsets any value already set on the trigger def
- array with value sets to just that value

We don't need to validate the response just that we can properly generate the necessary input for each of those states.

rocktavious (approved):

PR #246, opened 2023-09-12
fix input param nullability for extendedTeamAcess

davidbloss:
Fixing a missed `!` to make `IdentifierInput` -> `IdentifierInput!`


rocktavious:
@davidbloss can we get a changelog entry for this since i shipped opslevel-go and the CLI last night we'll need to release a hotfix

rocktavious (approved):

PR #250, opened 2023-09-21
Db/add taggable resource interface

davidbloss:
Draft for now:
We expected similar behaviors and traits from these [TaggableResource](https://github.com/OpsLevel/opslevel-go/blob/main/enum.go#L710-L716)s. 

Putting these in a `TaggableResourceInterface` interface simplifies:
- Getting the specific type of the TaggableResource
- Getting the resource ID that the tag belongs to
- Getting the specific tag ID from the list of tags belonging to the TaggableResource

### DIFF
--- a/tags.go
+++ b/tags.go
@@ -7,6 +7,12 @@ const (
 	TagOwnerRepository TagOwner = "Repository"
 )
 
+type TaggableResourceInterface interface {
+	GetTag(ID) (*Tag, *Client)
+	ResourceId() ID
+	ResourceType() TaggableResource
+}
+
 type Tag struct {
 	Id    ID     `json:"id"`
 	Key   string `json:"key"`
--- a/.changes/unreleased/Feature-20230921-092515.yaml
+++ b/.changes/unreleased/Feature-20230921-092515.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: add interface for taggable resources
+time: 2023-09-21T09:25:15.839769-05:00
--- a/repository.go
+++ b/repository.go
@@ -106,6 +106,32 @@ type ServiceRepositoryUpdateInput struct {
 	DisplayName   string `json:"displayName,omitempty"`
 }
 
+func (r *Repository) GetTag(tagId ID, client *Client) *Tag {
+	if r == nil {
+		return nil
+	}
+	tags, err := r.GetTags(client, nil)
+	if err != nil {
+		fmt.Println(fmt.Errorf("Error getting tags: %s", err))
+		return nil
+	}
+	for _, tag := range tags.Nodes {
+		if tag.Id == tagId {
+			return &tag
+		}
+	}
+	fmt.Println(fmt.Errorf("Error getting tags: %s", err))
+	return nil
+}
+
+func (r *Repository) ResourceId() ID {
+	return r.Id
+}
+
+func (r *Repository) ResourceType() TaggableResource {
+	return TaggableResourceRepository
+}
+
 func (r *Repository) GetService(service ID, directory string) *ServiceRepository {
 	for _, edge := range r.Services.Edges {
 		for _, connection := range edge.ServiceRepositories {


taimoor-at-opslevel (approved):

system.go
```diff
@@ -28,6 +28,32 @@ type SystemInput struct {
...
+	if err != nil {
+		fmt.Println(fmt.Errorf("Error getting tags: %s", err))
+		return nil
+	}
+	for _, tag := range tags.Nodes {
+		if tag.Id == tagId {
+			return &tag
+		}
+	}
+	fmt.Println(fmt.Errorf("Error getting tags: %s", err))
```
rocktavious, line 45: the client library shouldn't have any output like this - it should go into an error if its an error otherwise return a sane default

system.go
```diff
@@ -28,6 +28,32 @@ type SystemInput struct {
 	Note        *string          `json:"note,omitempty"`
 }
 
+func (s *System) GetTag(tagId ID, client *Client) *Tag {
+	if s == nil {
+		return nil
+	}
+	tags, err := s.SystemId.Tags(client, nil)
+	if err != nil {
+		fmt.Println(fmt.Errorf("Error getting tags: %s", err))
```
rocktavious, line 37: same as below goes for here

rocktavious (changes requested):
Lets talk through the error handling in the GetTag functions

system.go
```diff
@@ -28,6 +28,32 @@ type SystemInput struct {
 	Note        *string          `json:"note,omitempty"`
 }
 
+func (s *System) GetTag(tagId ID, client *Client) *Tag {
```
rocktavious, line 31: We should also reverse the order of these arguments - in other golang tools the client and/or context for the client is usually the first argument.

system.go
```diff
@@ -28,6 +28,32 @@ type SystemInput struct {
...
+	tags, err := s.SystemId.Tags(client, nil)
+	if err != nil {
+		fmt.Println(fmt.Errorf("Error getting tags: %s", err))
+		return nil
+	}
+	for _, tag := range tags.Nodes {
+		if tag.Id == tagId {
+			return &tag
+		}
+	}
```
rocktavious, line 44: Since we do this on every function we should refactor this to be a utitlity function on a `TagConnection` 

so you can just do `return tags.FindTag(tagId)`

system.go
```diff
@@ -28,6 +28,32 @@ type SystemInput struct {
 	Note        *string          `json:"note,omitempty"`
 }
 
+func (s *System) GetTag(tagId ID, client *Client) *Tag {
+	if s == nil {
```
rocktavious, line 32:
```suggestion
	if s.ID == nil {
```

davidbloss: I could add a `s.ID == ""` check here.
The aim of the `if s == nil` check is to catch the unlikely event that
`var mySystem *opslevel.System` is declared and the pointer is never given a value, but later used to call GetTag()

taimoor-at-opslevel: +1 for also a checking for `s.ID == ""`, it could lead to an annoying bug down the road.

taimoor-at-opslevel:
@davidbloss is adding support for `User` and `InfrastructureResource` also in scope for this PR?

davidbloss:
> @davidbloss is adding support for `User` and `InfrastructureResource` also in scope for this PR?

Not for this PR. Those resources need Tag support introduced in the first place with tests.

domain.go
```diff
@@ -26,6 +26,29 @@ type DomainInput struct {
 	Note        *string `json:"note,omitempty"`
 }
 
+func (d *Domain) GetTag(client *Client, tagId ID) (*Tag, error) {
+	if d == nil {
```
rocktavious, line 30: are you able to make this happen?  My understanding is if you do
```go
var myDomain *Domain
myDomain.GetTag()
```
you won't even get into this function at all you'll get a `GetTag() function doesn't exist on "nil" object` error

domain.go
```diff
@@ -26,6 +26,29 @@ type DomainInput struct {
...
+	tags, err := d.DomainId.Tags(client, nil)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting tags: %s", err)
+	}
+	tag, err := tags.GetTagById(tagId)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting tags: %s", err)
+	}
+	return tag, nil
+}
```
rocktavious, line 42: This can be simplified as
```suggestion
	return tags.GetTagById(tagId)
}
```
 unless you thing the extra `Error getting tags` is helpful to the end user somehow?

domain.go
davidbloss: Oh, [it's possible](https://go.dev/play/p/w-fpHemL7Zt)

domain.go
davidbloss: Wow.... this is a great call out 👏

taimoor-at-opslevel (approved):

domain.go
```diff
@@ -26,29 +26,41 @@ type DomainInput struct {
 	Note        *string `json:"note,omitempty"`
 }
 
-func (s *DomainId) ChildSystems(client *Client, variables *PayloadVariables) (*SystemConnection, error) {
+func (d *Domain) GetTags(client *Client, variables *PayloadVariables) (*TagConnection, error) {
+	return d.DomainId.Tags(client, variables)
```
rocktavious, line 30: We should put this on the lower composite object

```suggestion
func (d *DomainId) GetTags(client *Client, variables *PayloadVariables) (*TagConnection, error) {
	return d.Tags(client, variables)
```

system.go
```diff
@@ -28,6 +28,18 @@ type SystemInput struct {
 	Note        *string          `json:"note,omitempty"`
 }
 
+func (s *System) GetTags(client *Client, variables *PayloadVariables) (*TagConnection, error) {
+	return s.SystemId.Tags(client, variables)
```
rocktavious, line 32: same as domain

```suggestion
func (s *SystemId) GetTags(client *Client, variables *PayloadVariables) (*TagConnection, error) {
	return s.Tags(client, variables)
```

tags.go
```diff
@@ -59,6 +66,54 @@ type TagDeleteInput struct {
 	Id ID `json:"id"`
 }
 
+func (client *Client) GetTaggableResource(resourceType TaggableResource, identifier string) (TaggableResourceInterface, error) {
+	var err error
+	var taggableResource TaggableResourceInterface
+
+	if !slices.Contains(AllTaggableResource, string(resourceType)) {
+		return nil, fmt.Errorf("not a taggable resource type: %s" + string(resourceType))
+	}
```
rocktavious, line 75: couldn't this be acheived by using the `default` case in the below switch statement?

davidbloss: Why....... yes. Yes it could 🤯

rocktavious (approved):
🎉

PR #251, opened 2023-09-22
add tag key validation

rocktavious:


tags.go
```diff
@@ -49,6 +59,15 @@ type TagDeleteInput struct {
 	Id ID `json:"id"`
 }
 
+//#region Helpers
+
+func ValidateTagKey(key string) error {
```
davidbloss, line 64: What do you think about making this a `func (t *Tag) ValidateTagKey()...` method?
The `TagKeyRegex` and `TagKeyErrorMsg` vars could live in there too.

rocktavious: ou i like that idea!

rocktavious: this was not possible - we paired on it together

davidbloss (approved):
LGTM 🎉

PR #255, opened 2023-09-25
Add `Tags *TagConnection` to all taggable resources

taimoor-at-opslevel:
## Issue

https://github.com/OpsLevel/team-platform/issues/104

## Changelog

- [x] Add support for Domain
- [x] Add support for System
- [x] Add support for InfrastructureResource
- [x] Add support for User

All changes were changing the GraphQL query. My method of doing so was looking at the `exp` vs `got` output, getting the diff, and making the appropriate string replacement. 

All tests and fixtures have been updated to accommodate adding `*TagConnection`. 

## Tophatting

See https://github.com/OpsLevel/cli/pull/182


rocktavious (changes requested):
We should not do this change.

We should actually be moving the resources that still use `TagConnection` as part of the get or list calls to move do the way Domains, Systems and Infra resources does it.  Which is by having an explicit `domain.Tags()` method call so the initial graphql requests a lighter weight and only when you actually need the tags do you call `domain.Tags()`

for example if a customer did this
```
func ContrivedExample() []string {
    domains, err := client.ListDomains(nil)
    output := []string{}
    for _, domain := range domains {
        output = append(output, domain.Name)
    }
    return output
}
```
this would now be a much heavier weight graphql call when they don't even need the tags on a domain.

PR #263, opened 2023-10-03
[ADDED] transport option in client settings for rest client

### LINKED ISSUE #197: Allow setting a custom transport when instantiating clients
Hey 👋 

I might be missing it, but I can't see any way to provide a custom http transport to the client.

For people building integrations inside larger applications, it's normal to provide a custom transport to provide standardised logging/tracing/retrying into their third-party clients.

Would you consider adding another client.Option that supports configuring a base transport?

Thanks!

zakisk:
## Issues

This PR fixes #197
## Changelog

- [x] added http transport as client settings option rest client

### DIFF
--- a/client.go
+++ b/client.go
@@ -2,6 +2,7 @@ package opslevel
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
@@ -9,12 +10,13 @@ import (
 )
 
 type ClientSettings struct {
-	url      string
-	token    string
-	timeout  time.Duration
-	retries  int
-	headers  map[string]string
-	pageSize int // Only Used by GQL
+	url       string
+	token     string
+	timeout   time.Duration
+	retries   int
+	headers   map[string]string
+	pageSize  int // Only Used by GQL
+	transport http.RoundTripper
 }
 
 type Option func(*ClientSettings)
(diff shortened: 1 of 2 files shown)


rocktavious (changes requested):
@zakisk - This only sets the transport for the REST client which is only used by a few of our API features - IE Integration URLs for CECs, Deploy Events, etc

The bulk of our API is in graphql and there is a clientGQL.go file that sets up that client.  That file specifically make use a couple of wrapped transport to provide timeout, retry, backoff, etc.

For us to accept this MR we'd need a solution that works for both OR we'd need to somehow scope this only to the REST side if thats what you need the custom transport for.

zakisk:
@rocktavious yeah, I'll think a solution for both clients. Thank you

zakisk:
@rocktavious Hello, I would like to inquire if it's possible to add another transport in ClinetSettings for GraphQL client and will use the default [transport](https://github.com/hashicorp/go-retryablehttp/blob/571a88bc9c3b7c64575f0e9b0f646af1510f2c76/roundtripper.go#L19) from retryablehttp if user doesn't pass one? Does this solution appear to be viable and effective to you?

rocktavious:
@zakisk - I guess whats the root cause we are trying to fix.  If the root problem is that the REST client doesn't automatically support retry and backoff, would it be better to setup the retryablehttp transport for the REST client like we did with the GraphQL client?  

Maybe we make the default transport in the ClientSettings the retryablehttp one and feed it into both REST and GraphQL.  Then if the end user wants to override that its on them to provide a transport that supports retry, backoff, etc ?

Personally i would prefer to minimize as many variable code paths as possible so know the root cause we are trying to fix would be helpful.

rocktavious:
Closing due to staleness

PR #264, opened 2023-10-03
add Variables to TestRequest struct, update tests

davidbloss:
## Issues

[ #100](https://github.com/OpsLevel/team-platform/issues/100)

## Changelog

Add `Variables` to `TestRequest` struct. First step toward improving testing.

- [X] List your changes here
- [ ] Make a `changie` entry

## Tophatting

All tests still pass - `task test`

PR #265, opened 2023-10-03
make a enum gen workflow

rocktavious:
## Changelog

- [ ] Create a new workflow that helps people regenerate enum.go

## Tophatting

Have not tested this yet

### DIFF
--- a/check_has_owner.go
+++ b/check_has_owner.go
@@ -1,28 +1,28 @@
 package opslevel
 
 type ServiceOwnershipCheckFragment struct {
-	RequireContactMethod *bool                             `graphql:"requireContactMethod"`
-	ContactMethod        *ServiceOwnershipCheckContactType `graphql:"contactMethod"`
-	TeamTagKey           string                            `graphql:"tagKey"`
-	TeamTagPredicate     *Predicate                        `graphql:"tagPredicate"`
+	RequireContactMethod *bool        `graphql:"requireContactMethod"`
+	ContactMethod        *ContactType `graphql:"contactMethod"`
+	TeamTagKey           string       `graphql:"tagKey"`
+	TeamTagPredicate     *Predicate   `graphql:"tagPredicate"`
 }
 
 type CheckServiceOwnershipCreateInput struct {
 	CheckCreateInput
 
-	RequireContactMethod *bool                             `json:"requireContactMethod,omitempty"`
-	ContactMethod        *ServiceOwnershipCheckContactType `json:"contactMethod,omitempty"`
-	TeamTagKey           string                            `json:"tagKey,omitempty"`
-	TeamTagPredicate     *PredicateInput                   `json:"tagPredicate,omitempty"`
+	RequireContactMethod *bool           `json:"requireContactMethod,omitempty"`
+	ContactMethod        *ContactType    `json:"contactMethod,omitempty"`
+	TeamTagKey           string          `json:"tagKey,omitempty"`
+	TeamTagPredicate     *PredicateInput `json:"tagPredicate,omitempty"`
 }
 
 type CheckServiceOwnershipUpdateInput struct {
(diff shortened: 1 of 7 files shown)


check_has_owner.go
```diff
@@ -1,28 +1,28 @@
 package opslevel
 
 type ServiceOwnershipCheckFragment struct {
-	RequireContactMethod *bool                             `graphql:"requireContactMethod"`
-	ContactMethod        *ServiceOwnershipCheckContactType `graphql:"contactMethod"`
```
rocktavious, line 5: enum.go doesn't generate this enum and it seems like contact type is the same thing so consolidating this.

clientGQL.go
```diff
@@ -3,10 +3,11 @@ package opslevel
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/go-retryablehttp"
-	"github.com/hasura/go-graphql-client"
 	"net/http"
 	"strings"
+
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/hasura/go-graphql-client"
```
rocktavious, line 10: gofumpt changes that fail go vet

enum.go
```diff
@@ -298,27 +298,6 @@ var AllContactType = []string{
 	string(ContactTypeGitHub),
 }
 
-// TODO: This appears to be duplicative of the above and i'm not sure why we need it
```
rocktavious, line 301: basically - actioning this TODO - so that we can have enum.go be auto generated at the click of the button

enum.go
```diff
@@ -460,6 +439,7 @@ const (
 	PredicateKeyEnumTags           PredicateKeyEnum = "tags"            // Filter by `tags` field.
 	PredicateKeyEnumOwnerID        PredicateKeyEnum = "owner_id"        // Filter by `owner` field.
 	PredicateKeyEnumGroupIDs       PredicateKeyEnum = "group_ids"       // Filter by group hierarchy. Will return resources who's owner is in the group ancestry chain.
+	PredicateKeyEnumOwnerIDs       PredicateKeyEnum = "owner_ids"       // Filter by `owner` hierarchy. Will return resources who's owner is in the team ancestry chain.
```
rocktavious, line 442: hey look a new enum value 🎉

enum.go
```diff
@@ -787,14 +768,16 @@ var AllToolCategory = []string{
 type UserRole string
 
 const (
-	UserRoleUser  UserRole = "user"  // A regular user on the account.
-	UserRoleAdmin UserRole = "admin" // An administrator on the account.
+	UserRoleUser      UserRole = "user"       // A regular user on the account.
+	UserRoleAdmin     UserRole = "admin"      // An administrator on the account.
+	UserRoleBasicUser UserRole = "basic_user" // A basic user on the account with limited access.
```
rocktavious, line 773: This is due to the new RBAC feature flag.  I'm told that if a customer tries to use this when their feature flag is not on they will get an error message saying its not enabled on their account.  So this should be safe

clientGQL.go
davidbloss: The `test` github action related to this includes running `task lint` defined in the [Taskfile](https://github.com/OpsLevel/opslevel-go/blob/main/Taskfile.yml#L16). Specifically it runs `gofumpt -d .` with the `-d` flag printing a diff - gofumpt itself doesn't write any changes.

The failure is from `go vet ./...` 
```
Error: vet: ./infra_test.go:288:3: too few values in struct literal of type struct{Request string; Variables string; Response string}
```
caused by some changes made in a [recently merge PR](https://github.com/OpsLevel/opslevel-go/pull/264/files#diff-7fe36a091d93790fb1a943d0636ab952e254e8d8b090881cd19d57e3d4d503ebR70-R72)

davidbloss (approved):
LGTM

PR #270, opened 2023-10-10
backout the new relic integration fragment

rocktavious:
## Issues

More context can be found here - https://jk-labs.slack.com/archives/C01G8VB8ZCY/p1696967424643479

TLDR: An engineer who no longer works here added a fragment for a non-public API and this caused a bug with the Terraform provider when making AWS integrations.

Long term i think this integration API was suppose to be made public but it works differently from the AWS integration mutation (but it actually works the more correct way) so we'll need to get the catalog team to update the AWS mutations to work correctly & make the needed new relic mutations public if we want this back in.

For now this stages this change incase that doesn't get fixed.


davidbloss (approved):
LGTM

rocktavious:
This was fixed in the API so we don't need to merge this

Issue #276, opened 2023-10-18
deprecate ownerAlias in favor of ownerInput


rocktavious:
- [ ] on `ServiceCreateInput` add
```
ownerInput: IdentifierInput
```
- [ ] on `ServiceUpdateInput` add
```
ownerInput: IdentifierInput
```
- [ ] on `ServiceCreateInput` deprecate `ownerAlias`
- [ ] on `ServiceUpdateInput` deprecate `ownerAlias`

PR #281, opened 2023-10-20
deprecate ownerAlias

taimoor-at-opslevel:
## Issues

https://github.com/OpsLevel/opslevel-go/issues/276
## Changelog

- [x] Make owner match what's in API
- [x] Make a `changie` entry

## Tophatting

See CLI PR.


rocktavious (approved):
This is good - but lets keep this out till next week as this will break downstream CLI and Terraform releases that need to be made.

davidbloss (approved):
LGTM 🎉

PR #282, opened 2023-10-20
add AffectsOverallServiceLevels field to Scorecard and ScorecardInput

davidbloss:
## Issues

[#71](https://github.com/OpsLevel/team-platform/issues/71)

## Changelog

Add `AffectsOverallServiceLevels` field to `Scorecard` and `ScorecardInput` structs

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` passes all tests

### DIFF
--- a/common.go
+++ b/common.go
@@ -47,9 +47,7 @@ func NewString(value string) *string {
 // Bool is a helper routine that allocates a new bool value
 // to store v and returns a pointer to it.
 func Bool(v bool) *bool {
-	p := new(bool)
-	*p = v
-	return p
+	return &v
 }
 
 func HandleErrors(err error, errs []OpsLevelErrors) error {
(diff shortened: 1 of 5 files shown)


common.go
```diff
@@ -47,9 +47,7 @@ func NewString(value string) *string {
 // Bool is a helper routine that allocates a new bool value
 // to store v and returns a pointer to it.
 func Bool(v bool) *bool {
-	p := new(bool)
-	*p = v
-	return p
+	return &v
```
taimoor-at-opslevel, line 50: I'm confused about this change and the function in general, doesn't this have a different result than the comment above this function?

rocktavious: @taimoor-at-opslevel  So the problem is that you cannot do this
```
input := ScorecardInput{
  AffectsOverallServiceLevels: &true
}
```

So you have two ways to fix it - pass the bool value through a function to get it stored in a var and then return the address of the variable 
```
input := ScorecardInput{
  AffectsOverallServiceLevels: ol.Bool(true)
}
```
OR make a local var and give the address
```
myBool := true
input := ScorecardInput{
  AffectsOverallServiceLevels: &myBool
}
```

rocktavious: The way the function was originally written was unnecessary - its creates an unneeded pointer - the way David fixed it is more idomatic go

rocktavious (approved):

rocktavious:
Gonna merge this and fixup the doc string later

PR #110, opened 2023-01-31
completed custom action and trigger testing

matthewbrahms:

PR #286, opened 2023-10-26
update templated ids to be strings or key value pairs

davidbloss:
## Issues

[#110](https://github.com/OpsLevel/team-platform/issues/110)

In an effort to standardize templated test data, I started with `id`s and their hardcoded equivalents.
For starters:
- individual values like `Z2lkOi8vMTIzNDU2Nzg5OTg3NjU0MzIx` can be named as "value_string"
  - `{{ template "id_string" }}` -> this would need to be quoted in the calling string such as `"{{ template "id_string" }}"`
- Most other `"key": "value"` values can be named like "value" and called like `{{ template "id" }}`
  - `{ {{ template "id" }} }` demonstrates a JSON object with a single key: value pair inside.
  - Later we could write wrapper template functions to quote or `{ }` templates.

## Changelog

Primary updates made in `testdata/templates/common.tpl`
- `{{ template "id1_string" }}` resolves to `Z2lkOi8vMTIzNDU2Nzg5OTg3NjU0MzIx`
- `{{ template "id1" }}` resolves to `"id": "Z2lkOi8vMTIzNDU2Nzg5OTg3NjU0MzIx"`

- [X] List your changes here
- [ ] Make a `changie` entry, N/A test only updates

## Tophatting

`task test` passes all tests


taimoor-at-opslevel (approved):
Could not review every single line but LGTM.

PR #287, opened 2023-11-01
remove create, update Groups. deprecate Group field in Teams

davidbloss:
## Issues

[#124](https://github.com/OpsLevel/team-platform/issues/124)

### Removed
- `GroupInput` and `EntityOwnerGroup` structs
- `CreateGroup()` and `UpdateGroup()` methods from client
- `Group` field from `TeamCreateInput` and `TeamUpdateInput` structs

### Deprecated
- Added a comment above the `Group` field in `Team` as another reminder.

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` shows all tests passing

### DIFF
--- a/owner.go
+++ b/owner.go
@@ -1,41 +1,20 @@
 package opslevel
 
-type EntityOwnerGroup struct {
-	Alias string `json:"alias,omitempty" graphql:"groupAlias:alias"`
-	Id    ID     `json:"id"`
-}
-
 type EntityOwnerTeam struct {
 	Alias string `json:"alias,omitempty" graphql:"teamAlias:alias"`
 	Id    ID     `json:"id"`
 }
 
 type EntityOwner struct {
-	OnGroup EntityOwnerGroup `graphql:"... on Group"`
-	OnTeam  EntityOwnerTeam  `graphql:"... on Team"`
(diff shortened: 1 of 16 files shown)


owner.go
```diff
@@ -1,41 +1,20 @@
...
-	Alias string `json:"alias,omitempty" graphql:"groupAlias:alias"`
-	Id    ID     `json:"id"`
-}
-
 type EntityOwnerTeam struct {
 	Alias string `json:"alias,omitempty" graphql:"teamAlias:alias"`
 	Id    ID     `json:"id"`
 }
 
 type EntityOwner struct {
```
rocktavious, line 8: since this is no longer polymorphic - do we even still need it?

rocktavious (approved):
LGTM - only 1 question that we can log a ticket for and do later if needs be.

taimoor-at-opslevel (approved):
I am impartial on removing `EntityOwner` now vs doing it later.

owner.go
davidbloss: I thought about this but had not really looked into how EntityOwner might be used elsewhere or by customers.
If you think it's safe to remove this we can though

PR #291, opened 2023-11-06
Managed Aliases on Service

taimoor-at-opslevel:
## Issues

https://github.com/OpsLevel/team-platform/issues/135

## Changelog

- [x] Allow access to ManagedAliases on Service
- [x] Make a `changie` entry

## Tophatting

See https://github.com/OpsLevel/terraform-provider-opslevel/pull/146


group_test.go
```diff
@@ -140,13 +140,13 @@ func TestDescendantRepositories(t *testing.T) {
...
+		`"query GroupDescendantServicesList($after:String!$first:Int!$group:ID!){account{group(id: $group){descendantServices(after: $after, first: $first){nodes{apiDocumentPath,description,framework,htmlUrl,id,aliases,language,lifecycle{alias,description,id,index,name},managedAliases,name,owner{alias,id},preferredApiDocument{id,htmlUrl,source{... on ApiDocIntegration{id,name,type},... on ServiceRepository{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},timestamps{createdAt,updatedAt}},preferredApiDocumentSource,product,repos{edges{node{id,defaultAlias},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},{{ template "pagination_request" }},totalCount},tags{nodes{id,key,value},{{ template "pagination_request" }},totalCount},tier{alias,description,id,index,name},timestamps{createdAt,updatedAt},tools{nodes{category,categoryAlias,displayName,environment,id,url,service{id,aliases}},{{ template "pagination_request" }},totalCount}},{{ template "pagination_request" }},totalCount}}}}"`,
+		`{ {{ template "first_page_variables" }}, "group": "Z2lkOi8vb3BzbGV2ZWwvTmFtZXNwYWNlczo6R3JvdXAvMTI" }`,
```
rocktavious, line 144: whats the reason for changing this?

rocktavious (changes requested):
I think this is the incorrect change - we likely don't want to use the old aliases field but point it to the new `managedAliases` field.

This will also affect all the fixture changes you made where `aliases` becomes `managedAliases`

service.go
```diff
@@ -21,6 +21,7 @@ type Service struct {
 	ServiceId
 	Language                   string                       `json:"language,omitempty"`
 	Lifecycle                  Lifecycle                    `json:"lifecycle,omitempty"`
+	ManagedAliases             []string                     `json:"managedAliases,omitempty"`
```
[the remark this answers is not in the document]
taimoor-at-opslevel: @rocktavious This would prevent the user from being able to see the service's slug for example when doing `opslevel list services`.

rocktavious (approved):

group_test.go
taimoor-at-opslevel: I'm not really sure why I added this change. Put in a commit just now to reverse this. Best I can think of is that I did a bad rebase which caused these to be out of sync.

PR #292, opened 2023-11-07
Allow access to ManagedAliases on InfraResource

taimoor-at-opslevel:
## Issues

https://github.com/OpsLevel/team-platform/issues/112 -- prereq to supporting aliases on infra resources in terraform. 

Field was added to the graphql API.

## Changelog

- [x] Add the field
- [x] Make a `changie` entry


taimoor-at-opslevel:
**Closing this,** because it turns out that infra resources do not suffer from the problem where the aliases array contains the slug.

taimoor-at-opslevel:
**Closing this,** because it turns out that infra resources do not suffer from the problem where the aliases array contains the slug! We can just use the existing `Aliases`.

PR #295, opened 2023-11-09
bugfix(scalar.go): return nil rather than empty alias for NewIdentifi…

msobkowiak-olx:
NewIdentifier(string) will always produce a non-empty JSON structure, this will not respect the "omitempty" hint on various inputs which can be reproduced with Terraform Provider when provisioning new system:

```
resource "opslevel_system" "capability" {
  name        = var.capability_name
  description = var.capability_description
  note        = var.capability_note
}
```

Which yields:

```
Error: OpsLevel API Errors:
	- 'id' One of id or alias must be supplied
```

## Changelog
- modifieds scalar.go with a nil fallback if both Alias/ID are not supplied
- modifies tests to account for the change


taimoor-at-opslevel:
Thanks a lot for submitting this pull request! 

After reading your PR we have realized that `IdentifierInput` and `UserIdentifierInput` are essentially broken in our tools, which is causing `'id' One of id or alias must be supplied` to be thrown whenever a user tries to unset certain foreign keys on certain resources. 

We are working on a permanent fix for the next release. This will involve updating these input types to stop using `omitempty` on non-pointer fields across all of our tools. The fix will likely result in these types either being replaced altogether. We cannot include a fix in today's release because there are a lot of dependencies on these types.

msobkowiak-olx:
Thanks - you may keep it opened for a bit for now.

davidbloss:
👋 Hi @msobkowiak-olx, the underlying issue raised here has been fixed. Specifically, the [fields in the IdentifierInput struct needed to be pointers](https://github.com/OpsLevel/opslevel-go/pull/303/files#diff-84f2e004caa515d3375ee11a0d98ecfd5f1571abdf9a84865cac811c7ee021c4R34-R35) and we will be releasing this fix tomorrow.

Thank you for digging into this and bringing this to light 🙌

PR #301, opened 2023-11-10
make UserIdentifierInput fields pointers

davidbloss:
## Issues

[#145](https://github.com/OpsLevel/team-platform/issues/145) - partial fix

## Changelog

Make `UserIdentifierInput` fields pointers to enable expected `omitempty` json tag behavior

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` - all tests pass

### DIFF
--- a/team_test.go
+++ b/team_test.go
@@ -797,7 +797,7 @@ func TestTeamAddMemberhip(t *testing.T) {
 	team, _ := clientWithAlias.GetTeamWithAlias("example")
 	newMembership := ol.TeamMembershipUserInput{
 		Role: "user",
-		User: ol.UserIdentifierInput{Id: id1, Email: "kyle@opslevel.com"},
+		User: ol.UserIdentifierInput{Id: &id1, Email: ol.NewString("kyle@opslevel.com")},
 	}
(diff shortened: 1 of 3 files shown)


taimoor-at-opslevel (approved):
LGTM!

team_test.go
```diff
@@ -797,7 +797,7 @@ func TestTeamAddMemberhip(t *testing.T) {
 	team, _ := clientWithAlias.GetTeamWithAlias("example")
 	newMembership := ol.TeamMembershipUserInput{
 		Role: "user",
-		User: ol.UserIdentifierInput{Id: id1, Email: "kyle@opslevel.com"},
+		User: ol.UserIdentifierInput{Id: &id1, Email: ol.NewString("kyle@opslevel.com")},
```
rocktavious, line 800: I "think" this should probably actually be using `NewUserIdentifier("kyle@opslevel.com")`

davidbloss: Introducing `NewUserIdentifier()` here would make the line read as:
`User: NewUserIdentifier("kyle@opslevel.com"),`
but then we are getting a `UserIdentifier` without an `Id`

PR #308, opened 2023-11-15
Support for Property Definitions

taimoor-at-opslevel:
## Issues

https://github.com/OpsLevel/team-platform/issues/122

Relies on https://github.com/OpsLevel/opslevel-go/pull/317

## Changelog

- [x] Add CRUD functions for property definitions
- [x] Testing
- [x] Refactor for `JSONString`
- [x] Make a `changie` entry

## Tophatting

Tests should pass.

### DIFF
--- a/property.go
+++ b/property.go
@@ -4,7 +4,7 @@ import "fmt"
 
 type PropertyDefinitionInput struct {
 	Name   string `json:"name"`
-	Schema string `json:"schema"`
+	Schema JSON   `json:"schema" scalar:"true"`
 }
 
 type PropertyDefinition struct {
--- a/property.go
+++ b/property.go
@@ -4,14 +4,14 @@ import "fmt"
 
 type PropertyDefinitionInput struct {
 	Name   string `json:"name"`
-	Schema JSON   `json:"schema" scalar:"true"`
+	Schema JSON   `json:"schema"`
 }
 
 type PropertyDefinition struct {
 	Aliases []string `graphql:"aliases"`
 	Id      ID       `graphql:"id"`
 	Name    string   `graphql:"name"`
-	Schema  string   `graphql:"schema"`
+	Schema  JSON     `graphql:"schema" scalar:"true"`
 }
 
 type PropertyDefinitionConnection struct {
--- a/property.go
+++ b/property.go
@@ -45,7 +45,7 @@ func (client *Client) GetPropertyDefinition(input string) (*PropertyDefinition, 
 	}
 	err := client.Query(&q, v, WithName("PropertyDefinitionGet"))
 	if q.Account.Definition.Id == "" {
-		err = fmt.Errorf("PropertyDefinition with ID '%s' or Alias '%s' not found", string(*input.Id), *input.Alias)
+		err = fmt.Errorf("PropertyDefinition with ID or Alias matching '%s' not found", input)
 	}
 	return &q.Account.Definition, HandleErrors(err, nil)
 }
@@ -75,7 +75,7 @@ func (client *Client) ListPropertyDefinitions(variables *PayloadVariables) (Prop
 	return q.Account.Definitions, nil
 }
 
-func (client *Client) DeletePropertyDefinition(input IdentifierInput) error {
+func (client *Client) DeletePropertyDefinition(input string) error {
 	var m struct {
 		Payload struct {
 			DeletedAlias string           `graphql:"deletedAlias"`
--- a/property.go
+++ b/property.go
@@ -83,8 +83,9 @@ func (client *Client) DeletePropertyDefinition(input string) error {
 			Errors       []OpsLevelErrors `graphql:"errors"`
 		} `graphql:"propertyDefinitionDelete(resource: $input)"`
 	}
+	identifier := NewIdentifier(input)
 	v := PayloadVariables{
-		"input": input,
+		"input": identifier,
 	}
 	err := client.Mutate(&m, v, WithName("PropertyDefinitionDelete"))
 	return HandleErrors(err, m.Payload.Errors)
--- a/property.go
+++ b/property.go
@@ -11,7 +11,7 @@ type PropertyDefinition struct {
 	Aliases []string `graphql:"aliases"`
 	Id      ID       `graphql:"id"`
 	Name    string   `graphql:"name"`
-	Schema  string   `graphql:"schema"`
+	Schema  JSON     `graphql:"schema" scalar:"true"`
 }
 
 type PropertyDefinitionConnection struct {
--- a/property.go
+++ b/property.go
@@ -40,8 +40,9 @@ func (client *Client) GetPropertyDefinition(input string) (*PropertyDefinition, 
 			Definition PropertyDefinition `graphql:"propertyDefinition(input: $input)"`
 		}
 	}
+	identifier := NewIdentifier(input)
 	v := PayloadVariables{
-		"input": input,
+		"input": identifier,
 	}
 	err := client.Query(&q, v, WithName("PropertyDefinitionGet"))
 	if q.Account.Definition.Id == "" {
--- a/property.go
+++ b/property.go
@@ -17,7 +17,7 @@ type PropertyDefinition struct {
 type PropertyDefinitionConnection struct {
 	Nodes      []PropertyDefinition
 	PageInfo   PageInfo
-	TotalCount int
+	TotalCount int `graphql:"-"`
 }
 
 func (client *Client) CreatePropertyDefinition(input PropertyDefinitionInput) (*PropertyDefinition, error) {
@@ -34,7 +34,7 @@ func (client *Client) CreatePropertyDefinition(input PropertyDefinitionInput) (*
 	return &m.Payload.Definition, HandleErrors(err, m.Payload.Errors)
 }
 
-func (client *Client) GetPropertyDefinition(input IdentifierInput) (*PropertyDefinition, error) {
+func (client *Client) GetPropertyDefinition(input string) (*PropertyDefinition, error) {
 	var q struct {
 		Account struct {
 			Definition PropertyDefinition `graphql:"propertyDefinition(input: $input)"`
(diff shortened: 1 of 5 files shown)


property.go
```diff
@@ -0,0 +1,91 @@
+package opslevel
+
+import "fmt"
+
+type PropertyDefinitionInput struct {
+	Name   string `json:"name"`
+	Schema string `json:"schema"`
```
davidbloss, line 7: I think the type for `Schema` should be `JSON` with `json:"schema scalar:"true"`?
Similar to [InfrastructureResourceSchema](https://github.com/OpsLevel/opslevel-go/blob/main/infra.go#L10)

taimoor-at-opslevel: I am not sure, I am looking at infra_test.go and I don't see any instances of the `JSON` type used in the tests. @rocktavious should we use the `JSON` type or is a string OK?

rocktavious: So we need to ensure that whats gets sent to the API is "stringify" json so
```
{\"key\":\"value\"}
```

I'm pretty sure our JSON scaler type does that.  Additionally - they ended up making a new type in the API - `JSONString` IIRC - we may need to match that by making our own matching scalar type to ensure it works correctly.

For the end user using opslevel-go - i think it would be ideal for them to use `map[string]any`

property.go
```diff
@@ -0,0 +1,91 @@
...
+		} `graphql:"propertyDefinitionCreate(input: $input)"`
+	}
+	v := PayloadVariables{
+		"input": input,
+	}
+	err := client.Mutate(&m, v, WithName("PropertyDefinitionCreate"))
+	return &m.Payload.Definition, HandleErrors(err, m.Payload.Errors)
+}
+
+func (client *Client) GetPropertyDefinition(input IdentifierInput) (*PropertyDefinition, error) {
```
rocktavious, line 37:
```suggestion
func (client *Client) GetPropertyDefinition(input string) (*PropertyDefinition, error) {
```
Our interface should be flexible and take in a string and we try to push it to `IdentifierInput` using the construct that detects if its an ID or Alias

property.go
```diff
@@ -0,0 +1,91 @@
...
+	Aliases []string `graphql:"aliases"`
+	Id      ID       `graphql:"id"`
+	Name    string   `graphql:"name"`
+	Schema  string   `graphql:"schema"`
+}
+
+type PropertyDefinitionConnection struct {
+	Nodes      []PropertyDefinition
+	PageInfo   PageInfo
+	TotalCount int
```
rocktavious, line 20:
```suggestion
	TotalCount int `graphql:"-"`
```
We cannot rely on TotalCount from the API because they don't always calculate it the same due to filter and such.  So we should make this field not be requested and then build up our down total count value based on `len(Nodes)`

property.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	for q.Account.Definitions.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Definitions.PageInfo.End
+		resp, err := client.ListPropertyDefinitions(variables)
+		if err != nil {
+			return PropertyDefinitionConnection{}, err
+		}
+		q.Account.Definitions.Nodes = append(q.Account.Definitions.Nodes, resp.Nodes...)
+		q.Account.Definitions.PageInfo = resp.PageInfo
+		q.Account.Definitions.TotalCount += resp.TotalCount
```
rocktavious, line 73:
```suggestion
		q.Account.Definitions.TotalCount += len(q.Account.Definitions.Nodes)
```

property.go
```diff
@@ -0,0 +1,91 @@
...
+		Account struct {
+			Definition PropertyDefinition `graphql:"propertyDefinition(input: $input)"`
+		}
+	}
+	v := PayloadVariables{
+		"input": input,
+	}
+	err := client.Query(&q, v, WithName("PropertyDefinitionGet"))
+	if q.Account.Definition.Id == "" {
+		err = fmt.Errorf("PropertyDefinition with ID '%s' or Alias '%s' not found", string(*input.Id), *input.Alias)
```
rocktavious, line 48: changing the input to this function also simplifies this.

rocktavious (changes requested):
Needs a few tweaks before its ready to ship

property_test.go
```diff
@@ -0,0 +1,95 @@
...
+	ol "github.com/opslevel/opslevel-go/v2023"
+	"github.com/rocktavious/autopilot/v2023"
+)
+
+// TODO: template variable for reused schema
+// TODO: template variable for ID's
+// TODO: template variable for prop1,prop2,prop3
+
+func TestCreatePropertyDefinition(t *testing.T) {
+	// Arrange
```
davidbloss, line 15: Could use
```go
propertyDefInput := autopilot.Register[[]ol.PropertyDefinitionInput]("property_definition_input",
ol.PropertyDefinitionInput{
  Name:   "my-prop",
  Schema: "{\"$ref\":\"#/$defs/MyProp\",\"$defs\":{\"MyProp\":{\"properties\":{\"name\":{\"type\":\"string\",\"title\":\"the name\",\"description\":\"The name of a friend\",\"default\":\"alex\",\"examples\":[\"joe\",\"lucy\"]}},\"additionalProperties\":false,\"type\":\"object\",\"required\":[\"name\"]}}}",
})
```

Then `{{ template "property_definition_input" }}` can be used as the variables (second) arg in `NewTestRequest()`

and `propertyDefInput` (or whatever it's named) can be passed into `client.CreatePropertyDefinition()` below

davidbloss: Example of this found in [domain_test.go](https://github.com/OpsLevel/opslevel-go/blob/main/domain_test.go#L12-L18)

property_test.go
```diff
@@ -0,0 +1,103 @@
...
+	autopilot.Ok(t, err)
+	autopilot.Equals(t, "XXX", string(property.Id))
+	autopilot.Equals(t, "my-prop", property.Name)
+	autopilot.Equals(t, schema1, property.Schema)
+}
+
+func TestDeletePropertyDefinition(t *testing.T) {
+	// Arrange
+	testRequest := NewTestRequest(
+		`"mutation PropertyDefinitionDelete($input:IdentifierInput){propertyDefinitionDelete(resource: $input){deletedAlias,deletedId,errors{message,path}}}"`,
```
taimoor-at-opslevel, line 46: Need some help double checking here: should this be 
```suggestion
		`"mutation PropertyDefinitionDelete($input:IdentifierInput!){propertyDefinitionDelete(resource: $input){deletedAlias,deletedId,errors{message,path}}}"`,
```
?

property.go
```diff
@@ -0,0 +1,93 @@
...
+func (client *Client) GetPropertyDefinition(input string) (*PropertyDefinition, error) {
+	var q struct {
+		Account struct {
+			Definition PropertyDefinition `graphql:"propertyDefinition(input: $input)"`
+		}
+	}
+	identifier := *NewIdentifier(input)
+	v := PayloadVariables{
+		"input": identifier,
+	}
```
davidbloss, line 46:
```suggestion
        v := PayloadVariables{
            "input": *NewIdentifier(input),
        }
```
nit: This is more in line with how `PayloadVariables` is defined in `infra.go`, `system.go`, and a couple others

property.go
```diff
@@ -0,0 +1,93 @@
...
+		Payload struct {
+			DeletedAlias string           `graphql:"deletedAlias"`
+			DeletedID    ID               `graphql:"deletedId"`
+			Errors       []OpsLevelErrors `graphql:"errors"`
+		} `graphql:"propertyDefinitionDelete(resource: $input)"`
+	}
+	identifier := *NewIdentifier(input)
+	v := PayloadVariables{
+		"input": identifier,
+	}
```
davidbloss, line 90:
```suggestion
        v := PayloadVariables{
            "input": *NewIdentifier(input),
        }
```
Same as above.
nit: This is more in line with how `PayloadVariables` is defined in `infra.go`, `system.go`, and a couple others

property.go
```diff
@@ -0,0 +1,91 @@
...
+		Account struct {
+			Definition PropertyDefinition `graphql:"propertyDefinition(input: $input)"`
+		}
+	}
+	v := PayloadVariables{
+		"input": *NewIdentifier(input),
+	}
+	err := client.Query(&q, v, WithName("PropertyDefinitionGet"))
+	if q.Account.Definition.Id == "" {
+		err = fmt.Errorf("PropertyDefinition with ID or Alias matching '%s' not found", input)
```
davidbloss, line 48: Let's remove `or Alias` from the error message since it's only checking for the `Id`

rocktavious: no the input to the function can take an alias or id - thats what its saying.  This is all good.

property.go
```diff
@@ -0,0 +1,91 @@
...
+		q.Account.Definitions.TotalCount += len(q.Account.Definitions.Nodes)
+	}
+	return q.Account.Definitions, nil
+}
+
+func (client *Client) DeletePropertyDefinition(input string) error {
+	var m struct {
+		Payload struct {
+			DeletedAlias string           `graphql:"deletedAlias"`
+			DeletedID    ID               `graphql:"deletedId"`
```
rocktavious, line 82: since we don't return these fields from this function there is no reason they need to be in the response body.

taimoor-at-opslevel: Removed these from here and from the test.

davidbloss (approved):
LGTM 🎉

property.go
```diff
@@ -0,0 +1,89 @@
...
+type PropertyDefinitionInput struct {
+	Name   string     `json:"name"`
+	Schema JSONString `json:"schema"`
+}
+
+type PropertyDefinition struct {
+	Aliases []string `graphql:"aliases" json:"aliases"`
+	Id      ID       `graphql:"id" json:"id"`
+	Name    string   `graphql:"name" json:"name"`
+	Schema  JSON     `json:"schema" scalar:"true"` // Do not add graphql struct tag here
```
rocktavious, line 14: well you can add `graphql` struct tags if you neeed - you actually just need `scalar:"true"` because `JSON is a custom scalar type we wrote.

rocktavious (approved):

PR #310, opened 2023-11-16
remove RequestWithVariables func

davidbloss:
## Issues

[#117](https://github.com/OpsLevel/team-platform/issues/117)

## Changelog

Remove `func RequestWithVariables()` and put its logic in appropriate locations

- [X] List your changes here
- [ ] Make a `changie` entry, N/A test code only

## Tophatting

`task test` - all tests pass

### DIFF
--- a/clientGQL_test.go
+++ b/clientGQL_test.go
@@ -87,6 +87,14 @@ func NewTestRequest(request string, variables string, response string) TestReque
 	if !IsValidJson(testRequest.Variables) {
 		panic(fmt.Errorf("testRequest Variables is not valid json: '%s'", testRequest.Variables))
 	}
+	jsonRequestWithVariables := fmt.Sprintf(
+		`{"query": %s, "variables": %s}`,
+		testRequest.Request,
+		testRequest.Variables,
+	)
+	if !IsValidJson(jsonRequestWithVariables) {
+		panic(fmt.Errorf("testRequest Request with Variables is not valid json: '%s'", jsonRequestWithVariables))
+	}
 	if !IsValidJson(testRequest.Response) {
 		panic(fmt.Errorf("testRequest Response is not json: '%s'", testRequest.Response))
 	}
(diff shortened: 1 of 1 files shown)


clientGQL_test.go
```diff
@@ -87,6 +87,14 @@ func NewTestRequest(request string, variables string, response string) TestReque
 	if !IsValidJson(testRequest.Variables) {
 		panic(fmt.Errorf("testRequest Variables is not valid json: '%s'", testRequest.Variables))
 	}
+	jsonRequestWithVariables := fmt.Sprintf(
```
rocktavious, line 90: hmm - my initial hope with this was that we wouldn't need to do this.  Maybe it requires a change to autopilot first but i'd love to just use 
```
testRequest := TestRequest{
		Request:   Templated(request),
		Variables: Templated(variables),
		Response:  Templated(response),
	}
	```
and not have to do any fo the `fmt.Sprintf(` stuff

File: clientGQL_test.go [hunk 2]
```diff
@@ -166,7 +166,11 @@ func RegisterPaginatedEndpoint(t *testing.T, endpoint string, requests ...TestRe
 	url := fmt.Sprintf("/LOCAL_TESTING/%s", endpoint)
 	requestCount := 0
 	autopilot.Mux.HandleFunc(url, func(w http.ResponseWriter, r *http.Request) {
-		GraphQLQueryTemplatedValidation(t, requests[requestCount].RequestWithVariables())(r)
+		GraphQLQueryTemplatedValidation(t, fmt.Sprintf(
```
rocktavious, line 169: likewise - we need to adjust autopilot first so that we can pass in just `requests[requestCount]` and not have to do the `fmt.Sprintf(` stuff.

choosing to use the more structured way is the better thing long term.

davidbloss:
Continuing this work in [autopilot](https://github.com/rocktavious/autopilot/) ✨

PR #313, opened 2023-11-17
Db/generate api objects

davidbloss:
## Issues

[Generate Golang input structs](https://github.com/OpsLevel/team-platform/issues/164)

Generate Golang structs representing all public GraphQL `INPUT_OBJECTS`.
- The fields in these objects have `default:"example_value"` tags that will show up in `opslevel example <resource>`

### REMOVED:
- `Groups`
  - `group.go`
  - `group_test.go`
  - `testdata/templates/groups.tpl`
- Lots of deprecated member functions that have been marked for removal for a while.

### DEPRECATED:
- `NewString()`
- `NewInt()`
- `Bool()`

### ADDED:
- `input.go` has all (generated) `*Input` structs. Moved from various `*.go` files.
- `RefOf()` to replace `NewString()`, `NewInt()`, and `Bool()`
- `NewManualCheckFrequencyUpdateInput()` factory function

### UPDATES:
- Renamed `JSONString` to `JsonString` to match GraphQL API.
- Add `JSONSchema` type. Similar to `JSON`
- `PropertyDefinition.Schema` type is now `JSONSchema`, was `JSON`

- [X] List your changes here
- [x] Make a `changie` entry

## Tophatting

```go
task gen

# creates input.go
```


common.go
```diff
@@ -40,6 +36,10 @@ func NullString() *string {
 	return output
 }
 
+func NewEnum[T ~string](value T) *T {
```
davidbloss, line 39: Here `~string` means that `T` must have an underlying type that is a string.
Golang ref [here](https://go.dev/blog/deconstructing-type-parameters)

actions_test.go
```diff
@@ -134,11 +132,14 @@ func TestCreateTriggerDefinition(t *testing.T) {
...
 	trigger, err := client.CreateTriggerDefinition(ol.CustomActionsTriggerDefinitionCreateInput{
-		Name:        "Deploy Rollback",
-		Description: ol.NewString("Disables the Deploy Freeze"),
-		Action:      *newID,
-		Owner:       *newID,
-		Filter:      ol.NewID("987654321"),
+		Name:                   "Deploy Rollback",
+		AccessControl:          ol.NewEnum[ol.CustomActionsTriggerDefinitionAccessControlEnum](ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone),
+		Description:            ol.NewString("Disables the Deploy Freeze"),
+		ResponseTemplate:       ol.NewString(""),
```
taimoorgit, line 138: Passing a pointer to an empty string feels weird, why are we doing that in the test?

system.go
```diff
@@ -23,16 +23,41 @@ type SystemConnection struct {
...
 
-type SystemInput struct {
-	Name        *string          `json:"name,omitempty"`
-	Description *string          `json:"description,omitempty"`
-	Owner       *ID              `json:"ownerId,omitempty" yaml:"ownerId,omitempty"`
-	Parent      *IdentifierInput `json:"parent,omitempty" yaml:"parent,omitempty"`
-	Note        *string          `json:"note,omitempty"`
-}
-
 func (s *SystemId) GetTags(client *Client, variables *PayloadVariables) (*TagConnection, error) {
```
taimoorgit, line 26: `TagConnection` should always be the same, can we generate this function? Or better yet just not provide this function and rely on tags.go instead? Would cut down on circular dependencies too, I think.

team_test.go
```diff
@@ -762,14 +760,14 @@ func TestTeamAddContact(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation ContactCreate($input:ContactCreateInput!){contactCreate(input: $input){contact{address,displayName,id,type},errors{message,path}}}`,
-		`{"input": {"type":"slack", "address":"#general", "teamId": "{{ template "id1_string" }}" }}`,
+		`{"input": {"type":"slack", "address":"#general", "teamAlias": "{{ template "id1_string" }}" }}`,
```
taimoorgit, line 763: This should either remain as `teamId` or we shouldn't use this string template

property.go
```diff
@@ -2,12 +2,12 @@ package opslevel
 
 import "fmt"
 
-type PropertyDefinitionInput struct {
-	Name                  string                    `json:"name,omitempty" yaml:"name,omitempty" default:"Example Package Schema"`
-	Description           string                    `json:"description,omitempty" yaml:"description,omitempty" default:"Place description here"`
-	Schema                JSON                      `json:"schema,omitempty" yaml:"schema,omitempty" default:"{\"$schema\":\"https://json-schema.org/draft/2020-12/schema\",\"title\":\"Packages\",\"description\":\"A list of packages.\",\"type\":\"array\",\"items\":{\"type\":\"object\",\"properties\":{\"name\":{\"type\":\"string\"},\"version\":{\"type\":\"string\"},\"lock_file\":{\"type\":\"string\"},\"manager\":{\"type\":\"string\"}},\"required\":[\"name\",\"version\"]},\"minItems\":0,\"uniqueItems\":true}"`
-	PropertyDisplayStatus PropertyDisplayStatusEnum `json:"propertyDisplayStatus,omitempty" yaml:"propertyDisplayStatus,omitempty" default:"visible"`
-}
+// type PropertyDefinitionInput struct {
```
taimoorgit, line 5: Can we remove all this right now rather than comment it out?

infra.go
```diff
@@ -41,24 +41,12 @@ type InfrastructureResourceConnection struct {
 	TotalCount int `graphql:"-"`
 }
 
-type InfrastructureResourceSchemaInput struct {
-	Type string `json:"type" yaml:"type" default:"Database"`
-}
-
 type InfrastructureResourceProviderInput struct {
```
taimoorgit, line 44: Can this be removed right now too?

taimoorgit (changes requested):
Looks excellent so far

actions_test.go
taimoorgit: I figure you're doing this because the field was changed to a pointer just now and the original mocked response JSON passed an empty string,

In that case it can be considered out of scope to fix this. 👍

team.go
```diff
@@ -322,32 +255,19 @@ func (client *Client) AddMemberships(team *TeamId, memberships ...TeamMembership
...
 		} `graphql:"contactCreate(input: $input)"`
 	}
+	// TODO: there is no teamId field anymore
 	contactInput := ContactCreateInput{
 		Type:        contact.Type,
-		DisplayName: *contact.DisplayName,
+		DisplayName: contact.DisplayName,
 		Address:     contact.Address,
-	}
-	if IsID(team) {
```
davidbloss, line 347: This is being added back...

team_test.go
davidbloss: Great catch!! There was a related bug here and it's just been fixed.

common.go
```diff
@@ -40,13 +36,7 @@ func NullString() *string {
 	return output
 }
 
-func NewString(value string) *string {
```
rocktavious, line 43: Should we keep this around - mark them as deprecated and under the hood of them use `RefOf`

This will give opslevel-go upgraders a warning about switching off this for this new paradigm that will be just introduced in the 2024 release?

dependencies_test.go
```diff
@@ -17,11 +17,11 @@ func TestCreateServiceDependency(t *testing.T) {
 	client := BestTestClient(t, "serviceDependencyCreate", testRequest)
 	// Act
 	result, err := client.CreateServiceDependency(ol.ServiceDependencyCreateInput{
-		Key: ol.ServiceDependencyKey{
-			Service:   *ol.NewIdentifier("example_2"),
-			DependsOn: *ol.NewIdentifier("example_3"),
+		DependencyKey: ol.ServiceDependencyKey{
```
rocktavious, line 20: WTF - why are these optional.  🤦 our API is soo terrible sometimes.  functionality wise you need both 😮‍💨

enum.go
```diff
@@ -812,16 +868,14 @@ var AllToolCategory = []string{
 type UserRole string
 
 const (
-	UserRoleUser      UserRole = "user"       // A regular user on the account.
-	UserRoleAdmin     UserRole = "admin"      // An administrator on the account.
-	UserRoleBasicUser UserRole = "basic_user" // A basic user on the account with limited access.
+	UserRoleUser  UserRole = "user"  // A regular user on the account.
+	UserRoleAdmin UserRole = "admin" // An administrator on the account.
```
rocktavious, line 872: please undo this.

enum.go
```diff
@@ -812,16 +868,14 @@ var AllToolCategory = []string{
...
 )
 
 // All UserRole as []string
 var AllUserRole = []string{
 	string(UserRoleUser),
 	string(UserRoleAdmin),
-	string(UserRoleBasicUser),
```
rocktavious, line 824: same as above

infra.go
```diff
@@ -119,19 +101,20 @@ func (i *InfrastructureResource) ResourceType() TaggableResource {
 }
 
 func (client *Client) CreateInfrastructure(input InfraInput) (*InfrastructureResource, error) {
+	data := JSON(input.Data)
 	i := InfrastructureResourceInput{
 		Schema: &InfrastructureResourceSchemaInput{Type: input.Schema},
-		Data:   input.Data,
+		Data:   &data,
```
rocktavious, line 107: why can we not use `&input.Data` here?

json.go
```diff
@@ -38,20 +38,20 @@ func (s JSON) MarshalJSON() ([]byte, error) {
 	return []byte(strconv.Quote(string(b))), err
 }
 
-// JSONString is a specialized input type to support serialization to JSON for input to graphql
-type JSONString string
+// JsonString is a specialized input type to support serialization to JSON for input to graphql
+type JsonString string
```
rocktavious, line 42: make sure this makes it into a changie entry

integration_test.go
```diff
@@ -47,7 +47,7 @@ func TestCreateNewRelicIntegration(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation NewRelicIntegrationCreate($input:NewRelicIntegrationInput!){newRelicIntegrationCreate(input: $input){integration{id,name,type,createdAt,installedAt,... on AwsIntegration{iamRole,externalId,awsTagsOverrideOwnership,ownershipTagKeys},... on NewRelicIntegration{baseUrl,accountKey}},errors{message,path}}}`,
-		`{ "input": { "apiKey": "123456789", "baseUrl": "https://api.newrelic.com/graphql", "accountKey": "XXXX" }}`,
```
rocktavious, line 50: why is this being removed?

common.go
taimoorgit: @rocktavious yes I think that's a good idea. To be extra annoying we could even `log.Debug()` (or info? that might break stuff) about it. I don't think `ref.Of()` is something that comes to mind intuitively and would be a pretty big shock.

dependencies_test.go
taimoorgit: We could add an exception in the generator to force that :trollface:

infra.go
taimoorgit: There's a hardcoded rule in gen.go where `Data` always marshals to `JSON`.

integration_test.go
davidbloss: Humans wrote this 😂 
Also this is correct, we talked

dependencies_test.go
rocktavious: Na thats ok just realizing how bad our API is with optional values.

.changes/unreleased/Bugfix-20240105-134240.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Bugfix
```
taimoorgit, line 1:
```suggestion
kind: Refactor
```

taimoorgit (changes requested):
🤠 a couple things:

1. Can we change `JsonString` to translate into `JSONString` in gen.go? It is really, **_REALLY_** painful to mix capitalization between `JSONString` and `JSONSchema`. I'm in favour of doing this even if there is a capitalization mismatch with the API. 
2. We should remove any mentions of `default:""` tags in gen.go, since they can cause bugs. I don't think opslevel-go should be using `default:""` tags anywhere at all since those tags are consumed by other applications. [More on this here](https://github.com/OpsLevel/team-platform/issues/174)

json.go
```diff
@@ -8,7 +8,38 @@ import (
...
+
+func NewJSONSchema(data string) JSONSchema {
+	result := make(JSONSchema)
+	if err := json.Unmarshal([]byte(data), &result); err != nil {
+		panic(err)
+	}
+	return result
+}
+
+func (s JSONSchema) ToJSON() string {
```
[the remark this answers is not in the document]
rocktavious: We shouldn't use `String` as that implements the `stringer` interface and I don't think we want that so maybe `AsString()`?

service.go
```diff
@@ -732,7 +685,7 @@ func (client *Client) DeleteService(input ServiceDeleteInput) error {
 
 func (client *Client) DeleteServiceWithAlias(alias string) error {
```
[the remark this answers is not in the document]
rocktavious: So should only do this once we fix the TODO on line 670

taimoorgit (approved):

rocktavious (approved):

PR #315, opened 2023-11-24
Fix bug where service fields are not being updated

taimoor-at-opslevel:
Used here: https://github.com/OpsLevel/terraform-provider-opslevel/pull/160

### DIFF
--- a/service.go
+++ b/service.go
@@ -61,16 +61,15 @@ type ServiceCreateInput struct {
 }
 
 type ServiceUpdateInput struct {
-	Id          ID               `json:"id,omitempty"`
-	Alias       string           `json:"alias,omitempty"`
-	Name        string           `json:"name,omitempty"`
-	Product     string           `json:"product,omitempty"`
-	Description string           `json:"description,omitempty"`
-	Language    string           `json:"language,omitempty"`
-	Framework   string           `json:"framework,omitempty"`
-	Tier        string           `json:"tierAlias,omitempty"`
-	Owner       *IdentifierInput `json:"ownerInput,omitempty"`
-	Lifecycle   string           `json:"lifecycleAlias,omitempty"`
+	Id          *ID              `json:"id"`
+	Name        *string          `json:"name"`
+	Product     *string          `json:"product"`
+	Description *string          `json:"description"`
+	Language    *string          `json:"language"`
+	Framework   *string          `json:"framework"`
+	Tier        *string          `json:"tierAlias"`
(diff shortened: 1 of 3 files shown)


service.go
```diff
@@ -61,16 +61,15 @@ type ServiceCreateInput struct {
...
-	Owner       *IdentifierInput `json:"ownerInput,omitempty"`
-	Lifecycle   string           `json:"lifecycleAlias,omitempty"`
+	Id          *ID              `json:"id"`
+	Name        *string          `json:"name"`
+	Product     *string          `json:"product"`
+	Description *string          `json:"description"`
+	Language    *string          `json:"language"`
+	Framework   *string          `json:"framework"`
+	Tier        *string          `json:"tierAlias"`
+	Owner       *IdentifierInput `json:"ownerInput"`
```
rocktavious, line 71: i'm 99% sure you still need the `omitempty` because otherwise during "update" if you didn't specify the Language field but it was set this will wipe it out.

You should write a test to ensure you can generate an input json where some of the fields like `language` are missing from the input.

then the way to "unset" a value is to pass a pointer to an intialized string var that is "empty"

IE
```go
lang := ""
input := ServiceUpdateInput{
    Language: &lang,
}
```

This should generate the following json which will be devoid of all the other fields that were not set.

```json
{"language": ""}
```

rocktavious (changes requested):
This needs more tests to ensure you only "unset" the values you want to "unset"

service.go
taimoor-at-opslevel: If we use `omitempty` on a field, what's the best way for the user to be able to unset a special type? For example the `owner` of a Service. If we use `omitempty` with a pointer setting it to `nil` would not work.

If we have boolean or integer types in graphQL, the default values for those would be `false` and `0` respectively. I don't know what the way would be to set those specific values without the API thinking we want to unset them. 

My attempt to fix that was to treat updating services as a `PUT` rather than a `PATCH`, so the input struct should contain all information for the Service.

taimoor-at-opslevel:
> This needs more tests to ensure you only "unset" the values you want to "unset"

New changes I pushed does this by adding a new type `NullableInputString`. `omitempty` is re-added to `ServiceUpdateInput`, so fields can remain unchanged or set to `null OR "" OR "some non-empty string"`

This change would require refactoring for anyone using the new version as well as us updating the CLI, TF provider. It should be applied outside of Services as well.

PR #316, opened 2023-12-01
Bugfix for service fields not being unsettable

rocktavious:
The needed changes to fix a bug where service fields were not "unset"-able during service update

## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

The tests show if you pass in a new string pointer the field shows up in the input as being set to an empty string.


taimoorgit:
Closing this - cannot be done without a major change to how structs are defined and that work is being overwritten by gen.go changes in https://github.com/OpsLevel/opslevel-go/pull/313

PR #319, opened 2023-12-04
run changie workflow only if go files updated. doc,test only updates …

davidbloss:
## Issues

Have changie workflow check for changie logs only when actual `*.go` files are updated. Github workflow, docs, and testdata/test only updates do not need changie logs. Dependabot can't merge Dependabot PRs with changie workflows failing.

## Changelog

- [X] List your changes here
- [ ] Make a `changie` entry, N/A github workflow update

### DIFF
--- a/.github/workflows/changie-gen.yml
+++ b/.github/workflows/changie-gen.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     # catch when the PR is opened with the label or when the label is added
     types: [labeled]
+    paths:
+      - '**.go'
+      - '!**_test.go'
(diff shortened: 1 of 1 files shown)


taimoor-at-opslevel (approved):

rocktavious (changes requested):
The purpose of this workflow is to create change entries when dependabot opens a PR.  Dependabot only changes the `go.mod` file so this change renders this workflow useless.

What is the root problem you are trying to solve?

davidbloss:
nvm :)

PR #332, opened 2023-12-12
fix typos in default struct tags

davidbloss:
## Changelog
Fix typos. `defaults:""` should be `default:""`
The default for JSONString needs loads of these `\` 👀 

- [X] List your changes here
- [ ] Make a `changie` entry, N/A typo fixes

### DIFF
--- a/property.go
+++ b/property.go
@@ -3,15 +3,15 @@ package opslevel
 import "fmt"
 
 type PropertyDefinitionInput struct {
-	Name   string     `json:"name" defaults:"John Doe"`
-	Schema JSONString `json:"schema" defaults:"{\"$schema\":\"https://json-schema.org/draft/2020-12/schema\",\"title\":\"Packages\",\"description\":\"A list of packages.\",\"type\":\"array\",\"items\":{\"type\":\"object\",\"properties\":{\"name\":{\"type\":\"string\"},\"version\":{\"type\":\"string\"},\"lock_file\":{\"type\":\"string\"},\"manager\":{\"type\":\"string\"},},\"required\":[\"name\",\"version\"],},\"minItems\":0,\"uniqueItems\":true}"`
+	Name   string     `json:"name" default:"John Doe"`
+	Schema JSONString `json:"schema,string" default:"\"{\\\"$schema\\\":\\\"https://json-schema.org/draft/2020-12/schema\\\",\\\"title\\\":\\\"Packages\\\",\\\"description\\\":\\\"A list of packages.\\\",\\\"type\\\":\\\"array\\\",\\\"items\\\":{\\\"type\\\":\\\"object\\\",\\\"properties\\\":{\\\"name\\\":{\\\"type\\\":\\\"string\\\"},\\\"version\\\":{\\\"type\\\":\\\"string\\\"},\\\"lock_file\\\":{\\\"type\\\":\\\"string\\\"},\\\"manager\\\":{\\\"type\\\":\\\"string\\\"},},\\\"required\\\":[\\\"name\\\",\\\"version\\\"],},\\\"minItems\\\":0,\\\"uniqueItems\\\":true}\""`
 }
 
 type PropertyDefinition struct {
-	Aliases []string `graphql:"aliases" json:"aliases" defaults:"-"`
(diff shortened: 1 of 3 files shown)


property.go
```diff
@@ -3,15 +3,15 @@ package opslevel
...
 
 type PropertyDefinition struct {
-	Aliases []string `graphql:"aliases" json:"aliases" defaults:"-"`
-	Id      ID       `graphql:"id" json:"id" defaults:"-"`
-	Name    string   `graphql:"name" json:"name" defaults:"-"`
-	Schema  JSON     `json:"schema" scalar:"true" defaults:"-"` // Do not add graphql struct tag here
+	Aliases []string `graphql:"aliases" json:"aliases" default:"-"`
+	Id      ID       `graphql:"id" json:"id" default:"-"`
+	Name    string   `graphql:"name" json:"name" default:"-"`
+	Schema  JSON     `json:"schema" scalar:"true" default:"-"` // Do not add graphql struct tag here
```
taimoor-at-opslevel, line 14: Blocking

Can we remove this comment too at the end of the line too? It shows up in the godoc for the `Schema` field which is unnecessary and confusing.

user.go
```diff
@@ -30,14 +30,14 @@ type UserConnection struct {
 }
 
 type UserIdentifierInput struct {
-	Id    *ID     `graphql:"id" json:"id,omitempty" defaults:"-"`
-	Email *string `graphql:"email" json:"email,omitempty" defaults:"-"`
+	Id    *ID     `graphql:"id" json:"id,omitempty" default:"-"`
```
taimoor-at-opslevel, line 33: Blocking

Question -  can we just omit the `default` tag if there's no default to set, instead of setting it to `"-"`?

property.go
```diff
@@ -3,15 +3,15 @@ package opslevel
 import "fmt"
 
 type PropertyDefinitionInput struct {
-	Name   string     `json:"name" defaults:"John Doe"`
-	Schema JSONString `json:"schema" defaults:"{\"$schema\":\"https://json-schema.org/draft/2020-12/schema\",\"title\":\"Packages\",\"description\":\"A list of packages.\",\"type\":\"array\",\"items\":{\"type\":\"object\",\"properties\":{\"name\":{\"type\":\"string\"},\"version\":{\"type\":\"string\"},\"lock_file\":{\"type\":\"string\"},\"manager\":{\"type\":\"string\"},},\"required\":[\"name\",\"version\"],},\"minItems\":0,\"uniqueItems\":true}"`
+	Name   string     `json:"name" default:"John Doe"`
+	Schema JSONString `json:"schema,string" default:"\"{\\\"$schema\\\":\\\"https://json-schema.org/draft/2020-12/schema\\\",\\\"title\\\":\\\"Packages\\\",\\\"description\\\":\\\"A list of packages.\\\",\\\"type\\\":\\\"array\\\",\\\"items\\\":{\\\"type\\\":\\\"object\\\",\\\"properties\\\":{\\\"name\\\":{\\\"type\\\":\\\"string\\\"},\\\"version\\\":{\\\"type\\\":\\\"string\\\"},\\\"lock_file\\\":{\\\"type\\\":\\\"string\\\"},\\\"manager\\\":{\\\"type\\\":\\\"string\\\"},},\\\"required\\\":[\\\"name\\\",\\\"version\\\"],},\\\"minItems\\\":0,\\\"uniqueItems\\\":true}\""`
```
taimoor-at-opslevel, line 7: Non-blocking

Is it possible for us to use the the JSON/JSONString type here combined with `fmt.Sprintf()` to set the tags? It would be easier to read.

taimoor-at-opslevel (changes requested):

property.go
davidbloss: We can not. The golang spec says the struct [Tag](https://go.dev/ref/spec#Tag) must be a [sting literal](https://go.dev/ref/spec#string_lit). Basic string concatenating doesn't work either

user.go
davidbloss: Yes it looks like it behaves the same way

taimoor-at-opslevel (approved):

PR #337, opened 2023-12-18
Assign Properties on a Service

taimoor-at-opslevel:
## Issues

https://github.com/OpsLevel/team-platform/issues/123

## Changelog

- [x] Allow getting a property with `owner` and `definition`
- [x] Assign a property
- [x] Unassign a property
- [x] On a service, be able to do the `Service.properties` query like in graphql.
- [x] NEW - Handle case where `Property.Value` can be null (possible according to schema)
- [x] Make a `changie` entry

## Tophatting

Unit tests should pass/tested in graphiql.

### DIFF
--- a/owner.go
+++ b/owner.go
@@ -23,3 +23,15 @@ func (s *EntityOwnerTeam) AsTeam() TeamId {
 		Id:    s.Id,
 	}
 }
+
+type EntityOwnerService struct {
+	OnService ServiceId `graphql:"... on Service"`
+}
+
+func (s *EntityOwnerService) Aliases() []string {
+	return s.OnService.Aliases
+}
+
+func (s *EntityOwnerService) Id() ID {
+	return s.OnService.Id
+}
--- a/property.go
+++ b/property.go
@@ -45,13 +45,8 @@ type Property struct {
 	Value            JSONString           `graphql:"value"`
 }
 
-type ServicePropertiesEdge struct {
-	Cursor string    `graphql:"cursor"`
-	Node   *Property `graphql:"node"`
-}
-
 type ServicePropertiesConnection struct {
-	Edges      []ServicePropertiesEdge
+	Nodes      []Property
 	PageInfo   PageInfo
 	TotalCount int `graphql:"-"`
 }
@@ -206,7 +201,7 @@ func (service *Service) GetProperties(client *Client, variables *PayloadVariable
 	if service.Properties == nil {
 		service.Properties = &ServicePropertiesConnection{}
 	}
-	service.Properties.Edges = append(service.Properties.Edges, q.Account.Service.Properties.Edges...)
+	service.Properties.Nodes = append(service.Properties.Nodes, q.Account.Service.Properties.Nodes...)
 	service.Properties.PageInfo = q.Account.Service.Properties.PageInfo
 	for service.Properties.PageInfo.HasNextPage {
 		(*variables)["after"] = service.Properties.PageInfo.End
--- a/property.go
+++ b/property.go
@@ -151,7 +151,7 @@ func (client *Client) GetProperty(owner string, definition string) (*Property, e
 	}
 	err := client.Query(&q, v, WithName("PropertyGet"))
 	if q.Account.Property.Definition.Id == "" {
-		err = fmt.Errorf("Property with ID or alias matching '%s' on Service with ID or alias matching '%s' not found", owner, definition)
+		err = fmt.Errorf("Property with ID or alias matching '%s' on Service with ID or alias matching '%s' not found", definition, owner)
 	}
 	return &q.Account.Property, HandleErrors(err, nil)
 }
--- a/property.go
+++ b/property.go
@@ -145,9 +145,6 @@ func (client *Client) GetProperty(owner string, definition string) (*Property, e
 		"definition": *NewIdentifier(definition),
 	}
 	err := client.Query(&q, v, WithName("PropertyGet"))
-	if q.Account.Property.Definition.Id == "" {
-		err = fmt.Errorf("Property with ID or alias matching '%s' on Service with ID or alias matching '%s' not found", definition, owner)
-	}
 	return &q.Account.Property, HandleErrors(err, nil)
 }
 
--- a/property.go
+++ b/property.go
@@ -32,10 +32,10 @@ type PropertyDefinitionId struct {
 }
 
 type PropertyInput struct {
-	Owner         IdentifierInput `json:"owner"`
-	Definition    IdentifierInput `json:"definition"`
-	Value         JSONString      `json:"value"`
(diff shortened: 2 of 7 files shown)


property.go
```diff
@@ -20,6 +20,33 @@ type PropertyDefinitionConnection struct {
...
+	Value         JSONString      `json:"value"`
+	RunValidation *bool           `json:"runValidation,omitempty"`
+}
+
+// Treating Definition (PropertyDefinition) as an IdentifierInput
+// Treating Owner (... on Service) as an IdentifierInput
+type Property struct {
+	Definition       IdentifierInput  `graphql:"definition"`
+	Owner            IdentifierInput  `graphql:"owner"`
+	ValidationErrors []OpsLevelErrors `graphql:"validationErrors"`
```
rocktavious, line 35: In the hackday today with foundations it looks like they actually have a new error type - we may need to look at it and see if its actually like our `OpsLevelErrors` struct or if we need to create a new type

property.go
```diff
@@ -20,6 +20,33 @@ type PropertyDefinitionConnection struct {
...
+	Value            JSONString       `graphql:"value"`
+}
+
+type ServicePropertiesEdge struct {
+	Cursor string    `graphql:"cursor"`
+	Node   *Property `graphql:"node"`
+}
+
+type ServicePropertiesConnection struct {
+	Edges      []ServicePropertiesEdge
```
rocktavious, line 45: we don't generally uses edges - we use nodes.  Additionally the connection type is usually defined inline and not as a separate struct.

property.go
```diff
@@ -20,6 +20,33 @@ type PropertyDefinitionConnection struct {
 	TotalCount int `graphql:"-"`
 }
 
+type PropertyInput struct {
+	Owner         IdentifierInput `json:"owner"`
+	Definition    IdentifierInput `json:"definition"`
+	Value         JSONString      `json:"value"`
+	RunValidation *bool           `json:"runValidation,omitempty"`
```
rocktavious, line 27: is this actually exposed in the public API??

property.go
```diff
@@ -87,3 +114,82 @@ func (client *Client) DeletePropertyDefinition(input string) error {
...
+			Property Property `graphql:"property(owner: $owner, definition: $definition)"`
+		}
+	}
+	v := PayloadVariables{
+		"owner":      *NewIdentifier(owner),
+		"definition": *NewIdentifier(definition),
+	}
+	err := client.Query(&q, v, WithName("PropertyGet"))
+	if q.Account.Property.Definition.Id == nil {
+		err = fmt.Errorf("Property with ID or alias matching '%s' on Service with ID or alias matching '%s' not found", owner, definition)
```
rocktavious, line 130: this error message is not correct / the string template vars are backwards i think.

property.go
taimoor-at-opslevel: Yes AFAICT
image

property.go
taimoor-at-opslevel: They are backwards, going to correct this.

property.go
taimoor-at-opslevel: [I used the same design as dependencies.go](https://github.com/OpsLevel/opslevel-go/blob/aced0c74dc751af84965c53e220f611a14e1f539/dependencies.go#L26-L29)

So I think that the connection type is required to be defined a struct in this file [so that it can be added to the `Service` object.](https://github.com/OpsLevel/opslevel-go/blob/aced0c74dc751af84965c53e220f611a14e1f539/service.go#L36)

I will try to refactor this to use Nodes instead of edges.

taimoor-at-opslevel: @rocktavious I have refactored this to use `Nodes []Property` instead of edges.

property.go
taimoor-at-opslevel: Talked to Foundations and they said that `runValidation` is expected to go public with the rest.

property.go
rocktavious: ahh the reason the use edges is because they put extra data on the edges - which i hate - so we should only do that if they did this.  Otherwise i prefer to use nodes generally.

rocktavious (approved):

owner.go
```diff
@@ -23,3 +23,15 @@ func (s *EntityOwnerTeam) AsTeam() TeamId {
...
+	OnService ServiceId `graphql:"... on Service"`
+}
+
+func (s *EntityOwnerService) Aliases() []string {
+	return s.OnService.Aliases
+}
+
+func (s *EntityOwnerService) Id() ID {
+	return s.OnService.Id
+}
```
rocktavious, line 37: I don't love this - theres probably a better way - but lets ship it for now because its not a one way door to fix.

property.go
```diff
@@ -108,3 +133,82 @@ func (client *Client) DeletePropertyDefinition(input string) error {
...
+	service.Properties.Nodes = append(service.Properties.Nodes, q.Account.Service.Properties.Nodes...)
+	service.Properties.PageInfo = q.Account.Service.Properties.PageInfo
+	for service.Properties.PageInfo.HasNextPage {
+		(*variables)["after"] = service.Properties.PageInfo.End
+		_, err := service.GetProperties(client, variables)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return service.Properties, nil
```
rocktavious, line 213: I think we are missing calculating `TotalCount` here

taimoor-at-opslevel: Added code to update `TotalCount` in the latest commit.

property.go
```diff
@@ -26,6 +26,31 @@ type PropertyDefinitionConnection struct {
 	TotalCount int `graphql:"-"`
 }
 
+type PropertyDefinitionId struct {
+	Id      ID       `json:"id"`
+	Aliases []string `json:"aliases,omitempty"`
+}
+
+type PropertyInput struct {
```
davidbloss, line 34: Can you add `yaml: ` tags here too?

property.go
```diff
@@ -26,6 +26,31 @@ type PropertyDefinitionConnection struct {
...
+	Owner         IdentifierInput `json:"owner"`
+	Definition    IdentifierInput `json:"definition"`
+	Value         JSONString      `json:"value"`
+	RunValidation *bool           `json:"runValidation,omitempty"`
```
davidbloss, line 38:
```suggestion
	Owner         IdentifierInput `json:"owner" yaml:"owner" default:"owner"`
	Definition    IdentifierInput `json:"definition" yaml:"definition" default:"definition"`
	Value         JSONString      `json:"value" yaml:"value" default:"[\"any\",\"valid\",\"JSON\",1.23,{\"nested\",true}]"`
	RunValidation *bool           `json:"runValidation,omitempty" yaml:"runValidation,omitempty" default:"true"`
```

davidbloss (approved):
LGTM! 🎉

PR #343, opened 2023-12-27
Better nullability for IdentifierInput

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/158

Allow opslevel-go to pass `NewIdentifier()` (with no args) as `*IdentifierInput` which results in the API being sent `null`. This enables being able to unset optional fields in resource updates.

## Changelog

- [x] Change functions that are using the `IdentifierInput` as arguments to instead take `string`
- [x] Modify `JSONMarshal()` logic for `IdentifierInput` to match the logic of marshalling an empty `ID`
- [x] New test tables for `IdentifierInput`
- [x] Add missing `yaml` tags, which caused bugs in the CLI. When tophatting the CLI, the lack of yaml `omitempty` caused updates to not work correctly.
- [x] Make `changie` entries
- [x] Other tools (changes only need to be made on fields that are `*IdentifierInput`)
    - [x] CLI - https://github.com/OpsLevel/cli/pull/230
    - [x] Terraform - https://github.com/OpsLevel/terraform-provider-opslevel/pull/178
    - [x] Kubectl no changes needed.

## Tophatting

Unit tests pass.

### DIFF
--- a/actions_test.go
+++ b/actions_test.go
@@ -110,16 +110,14 @@ func TestDeleteWebhookAction(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation WebhookActionDelete($input:IdentifierInput!){customActionsWebhookActionDelete(resource: $input){errors{message,path}}}`,
-		`{"input":{"id": "123456789"}}`,
+		`{"input":{"alias": "123456789"}}`,
 		`{"data": {"customActionsWebhookActionDelete": { "errors": [] }}}`,
 	)
 
 	client := BestTestClient(t, "custom_actions/delete_action", testRequest)
 
 	// Act
-	err := client.DeleteWebhookAction(ol.IdentifierInput{
-		Id: newID,
-	})
+	err := client.DeleteWebhookAction("123456789")
 
 	// Assert
 	autopilot.Ok(t, err)
@@ -201,13 +199,13 @@ func TestGetTriggerDefinition(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`query TriggerDefinitionGet($input:IdentifierInput!){account{customActionsTriggerDefinition(input: $input){{ template "custom_actions_trigger_request" }}}}`,
-		`{"input":{"id":"123456789"}}`,
+		`{"input":{"alias":"123456789"}}`,
 		`{"data": {"account": { "customActionsTriggerDefinition": {{ template "custom_action_trigger2" }} }}}`,
 	)
 	client := BestTestClient(t, "custom_actions/get_trigger", testRequest)
 
 	// Act
-	trigger, err := client.GetTriggerDefinition(ol.IdentifierInput{Id: newID})
+	trigger, err := client.GetTriggerDefinition("123456789")
 	// Assert
 	autopilot.Ok(t, err)
 	autopilot.Equals(t, "Release", trigger.Name)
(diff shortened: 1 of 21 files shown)


actions_test.go
```diff
@@ -110,16 +110,14 @@ func TestDeleteWebhookAction(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation WebhookActionDelete($input:IdentifierInput!){customActionsWebhookActionDelete(resource: $input){errors{message,path}}}`,
-		`{"input":{"id": "123456789"}}`,
+		`{"input":{"alias": "123456789"}}`,
```
taimoorgit, line 113: TODO: replace "123456789" with an alias that makes sense.

scalar_test.go
```diff
@@ -88,51 +88,84 @@ func TestConstructMutationID(t *testing.T) {
...
-		IdInputConcreteOmitEmpty: *id2,
-		IdInputPointer:           id2,
-		IdInputPointerOmitEmpty:  id2,
-	}
-	case4 := IdentifierTester{
-		IdInputConcrete:          *id3,
-		IdInputConcreteOmitEmpty: *id3,
-		IdInputPointer:           id3,
-		IdInputPointerOmitEmpty:  id3,
+func TestMarshalIdentifiersOmitBehavior(t *testing.T) {
```
taimoorgit, line 127: This new test covers what is being tested in `TestMarshalIdentifier`, particularly the behaviour of `omitempty` on IdentifierInputs, so they can be used to unset fields during updates or leave fields unchanged during updates.

scalar.go
```diff
@@ -31,21 +32,48 @@ type Identifier struct {
...
+			id: NewID(value),
 		}
 	}
 	return &IdentifierInput{
-		Alias: NewString(value),
+		alias: NewString(value),
 	}
 }
 
+func EmptyIdentifier() *IdentifierInput {
```
taimoorgit, line 72: This is a necessary and special identifier that can be used to unset a field.

For example, you can pass `EmptyIdentifier()` in a `ServiceUpdateInput` to unset the `Owner` of a Service. 

Or, you can just not set a value for it and it the Service's owner will be untouched because of `omitempty`.

scalar.go
```diff
@@ -31,21 +32,48 @@ type Identifier struct {
 }
 
 type IdentifierInput struct {
```
taimoorgit, line 34: The identifier input's `id` and `alias` can now only be read through `input.Id()` and `input.Alias()`. As a result, `IdentifierInput` can no longer be mutated. 

This change is beneficial because:

1. It forces end users to use the `NewIdentifier()` function everywhere rather than instantiating identifiers on their own.
2. It prevents users from making malformed `IdentifierInputs` (for example one where there is no `id` and no `alias`, or the `id` and `alias` fields are swapped, or both `id` and `alias` are set.)

By locking down the `IdentifierInput` we can also create a special type to be used for `ServiceUpdateInput` to unset optional identifier fields, by editing the `MarshalJSON()` logic.

scalar.go
```diff
@@ -31,19 +32,36 @@ type Identifier struct {
...
+	var out string
+	if i.Id != nil {
+		out = fmt.Sprintf(`{"id":"%s"}`, string(*i.Id))
+	} else {
+		out = fmt.Sprintf(`{"alias":"%s"}`, string(*i.Alias))
+	}
+	return []byte(out), nil
+}
+
+func NewIdentifier(value ...string) *IdentifierInput {
```
davidbloss, line 52:
```go
func NewIdentifier(value ...string) *IdentifierInput {
	var output IdentifierInput
	if len(value) == 1 {
		if IsID(value[0]) {
			output.Id = NewID(value[0])
		} else {
			output.Alias = NewString(value[0])
		}
	}
	return &output
}
```
How about this?

davidbloss (approved):
LGTM! 🎉

PR #344, opened 2023-12-29
Add linter to validate struct tags

taimoorgit:
## Issues

Part of https://github.com/OpsLevel/team-platform/issues/158 we need to clean up the struct tags, which are sometimes incorrect or inconsistent. I ran into incorrect/missing struct tags causing issues in the CLI [recently](https://github.com/OpsLevel/opslevel-go/pull/343).

We can set rules for how tags need to be for input and non-input structs, pointer and non-pointer fields, and more. We could also add this to CI. 

## Changelog

- [ ] Add linter


taimoorgit:
Closing so we can instead focus on code generation. https://github.com/OpsLevel/team-platform/issues/163

PR #349, opened 2024-01-08
update README for v2024

davidbloss:
## Changelog

Update README after removing old functions like `AssignTagForId()` for `v2024`

- [X] List your changes here
- [ ] Make a `changie` entry, N/A README only update


rocktavious (approved):
Needs a few more changes to the "install" area but we can do that after 2024 is released

PR #350, opened 2024-01-08
Code improvements that will help with code gen

taimoorgit:
## Issues

These are code improvements that have nothing to do with code gen, but make code gen much easier.

The actual code gen stuff is being moved to another PR coming soon(tm).

## Changelog

~~- [x] Add `task parse` - a code parser that recognizes functions and resources.~~
- [x] Code improvements that increase consistency - see changie


gen.go
```diff
@@ -827,3 +836,131 @@ var templFuncMap = template.FuncMap{
...
+	}
+	defer file.Close()
+
+	root, err = parser.ParseFile(token.NewFileSet(), "./input.go", nil, parser.ParseComments)
+	if err != nil {
+		return err
+	}
+	// get names of simple DeleteInput structs that use (ID, string) as args
+	// generate New...DeleteInput(identifier string) functions from them.
+	ast.Inspect(root, func(n ast.Node) bool {
```
rocktavious, line 908: inventive 🎉

category_test.go
```diff
@@ -27,13 +27,13 @@ func TestCreateRubricCategory(t *testing.T) {
 func TestGetRubricCategory(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
-		`query CategoryGet($id:ID!){account{category(id: $id){id,name}}}`,
+		`query CategoryGet($id:String!){account{category(id: $id){id,name}}}`,
```
rocktavious, line 30: This is not a good change - we are changing the contract and i'm 99% this won't actually work in the API because its expected to be an ID type

category_test.go
```diff
@@ -43,13 +43,13 @@ func TestGetRubricCategory(t *testing.T) {
 func TestGetMissingRubricCategory(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
-		`query CategoryGet($id:ID!){account{category(id: $id){id,name}}}`,
+		`query CategoryGet($id:String!){account{category(id: $id){id,name}}}`,
```
rocktavious, line 46: same we shouldn't need to change this so this means our code is broken.

check_test.go
```diff
@@ -773,7 +773,7 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
...
...nnective,htmlUrl,id,name,predicates{key,keyData,type,value,caseSensitive}},id,level{alias,description,id,index,name},name,notes: rawNotes,owner{... on Team{alias,id}},type,... on AlertSourceUsageCheck{alertSourceNamePredicate{type,value},alertSourceType},... on CustomEventCheck{integration{id,name,type},passPending,resultMessage,serviceSelector,successCondition},... on HasRecentDeployCheck{days},... on ManualCheck{updateFrequency{startingDate,frequencyTimeScale,frequencyValue},updateRequiresComment},... on RepositoryFileCheck{directorySearch,filePaths,fileContentsPredicate{type,value},useAbsoluteRoot},... on RepositoryGrepCheck{directorySearch,filePaths,fileContentsPredicate{type,value}},... on RepositorySearchCheck{fileExtensions,fileContentsPredicate{type,value}},... on ServiceOwnershipCheck{requireContactMethod,contactMethod,tagKey,tagPredicate{type,value}},... on ServicePropertyCheck{serviceProperty,propertyValuePredicate{type,value}},... on TagDefinedCheck{tagKey,tagPredicate{type,value}},... on ToolUsageCheck{toolCategory,toolNamePredicate{type,value},toolUrlPredicate{type,value},environmentPredicate{type,value}},... on HasDocumentationCheck{documentType,documentSubtype}}}}`,
```
rocktavious, line 776: do we know whats changing here? i'm not seeing anything.

gen/config/alertsource.json
```diff
@@ -0,0 +1,13 @@
+{
```
rocktavious, line 1: what are these files?

input.go
```diff
@@ -1017,3 +1017,83 @@ type UsersFilterInput struct {
 	Arg  *string         `json:"arg,omitempty" yaml:"arg,omitempty" example:"example_arg"`      // Value to be filtered. (Optional.)
 	Type *BasicTypeEnum  `json:"type,omitempty" yaml:"type,omitempty" example:"does_not_equal"` // The operation applied to value on the field. (Optional.)
 }
+
+// create a new AlertSourceServiceDeleteInput by passing an ID string
+func NewAlertSourceServiceDeleteInput(value string) *AlertSourceServiceDeleteInput {
```
rocktavious, line 1022: Theres alot of lines of code in this file that are not covered by tests?  What the purpose of adding these uncovered lines?

repository_test.go
```diff
@@ -88,7 +88,7 @@ func TestGetRepositoryWithAlias(t *testing.T) {
...
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
-		`query RepositoryGet($repo:ID!){account{repository(id: $repo){archivedAt,createdOn,defaultAlias,defaultBranch,description,forked,htmlUrl,id,languages{name,usage},lastOwnerChangedAt,name,organization,owner{alias,id},private,repoKey,services{edges{atRoot,node{id,aliases},paths{href,path},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},{{ template "pagination_request" }},totalCount},tags{nodes{id,key,value},{{ template "pagination_request" }},totalCount},tier{alias,description,id,index,name},type,url,visible}}}`,
+		`query RepositoryGet($repo:String!){account{repository(id: $repo){archivedAt,createdOn,defaultAlias,defaultBranch,description,forked,htmlUrl,id,languages{name,usage},lastOwnerChangedAt,name,organization,owner{alias,id},private,repoKey,services{edges{atRoot,node{id,aliases},paths{href,path},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},{{ template "pagination_request" }},totalCount},tags{nodes{id,key,value},{{ template "pagination_request" }},totalCount},tier{alias,description,id,index,name},type,url,visible}}}`,
```
rocktavious, line 91: same as above - 99% sure this is a bad change that won't actually work in Graphiql.

rocktavious (changes requested):
Overall there are a number of questions and some concerns.

Especially because this makes us and users have to do this
```
	result, err := client.GetTeam(id1)
```
	
	becomes
	
```
	result, err := client.GetTeam(string(id1))
```

and this worsens the UX of the customer facing client functions

category_test.go
taimoorgit: @rocktavious you are correct and I have reverted this change. 

I originally changed this so that we can use `string` everywhere. While that's helpful if the graphql query supports an `ID` or an `Alias`, I realize now that it is not helpful for developers using our library in the case where the query ONLY accepts an `ID`. 

I have added logic to the linter part of `task parser` to acknowledge there's a situation where a `delete()` function can have args `string|ID`.

gen/config/alertsource.json
taimoorgit: @rocktavious these files are the output of `task parse` reading our opslevel-go functions and extracting all the details from them. A generator for a program like CLI can consume these .json files and use templates to generate code that uses opslevel-go.

repository_test.go
taimoorgit: @rocktavious you are correct, this was a mistake and has been reverted.

input.go
taimoorgit: @rocktavious originally the concern was that we would need these factory methods so that we could create these inputs using just a `string`. However that is not necessary so I'm reverting this change to input.go, as well as removing the code that generated this since it's also not necessary.

I do think we'll re-add something like this down the road when we start generating more opslevel-go code outside of just input.

gen/config/alertsource.json
rocktavious: long term the data should beable to be built up in memory from calling a function exposed by the library so if something in the CLI needs them it can just call the code machinery exposed by opslevel-go to do this.

Having a bunch of "data files" in a library is generally architecture smell to me unless they are test fixtures.

taimoorgit: @rocktavious The problem with the data not being serialized somewhere is that we would not be able to see changes in between versions.  That makes it much easier to track what changes in opslevel-go are affecting other repos and makes it easier to see what someone ended up changing in a pull request.  

I'm working on narrowing down the number of files generated since it is annoying. Working on a change now to make there only be 2 files: `generated.json` and `custom.json`, where `generated.json` is what the parser outputted and `custom.json` is full of any overrides we have set.

json.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // JSON is a specialized map[string]string to support proper graphql serialization
 type (
-	JSON       map[string]any
+	JSON       map[string]any // TODO: should this be marked as deprecated?
```
rocktavious, line 12: No its still a type in our API

property.go
```diff
@@ -92,19 +92,19 @@ func (client *Client) ListPropertyDefinitions(variables *PayloadVariables) (Prop
 		variables = client.InitialPageVariablesPointer()
 	}
 	if err := client.Query(&q, *variables, WithName("PropertyDefinitionList")); err != nil {
-		return PropertyDefinitionConnection{}, err
+		return &PropertyDefinitionConnection{}, err
```
rocktavious, line 95: After reading through all of these - i think if we are going to switch to returning a pointer these error checks should really return nil for the object.

```suggestion
		return nil, err
```

service_test.go
```diff
@@ -901,7 +901,7 @@ func TestDeleteService(t *testing.T) {
 	)
 	client := BestTestClient(t, "service/delete", testRequest)
 	// Act
-	err := client.DeleteService(ol.ServiceDeleteInput{Id: ol.NewID("Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS82NzQ3")})
+	err := client.DeleteService("Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS82NzQ3")
```
rocktavious, line 904: 😍

utils.go
```diff
@@ -1,11 +1,8 @@
 package opslevel
 
 import (
-	"encoding/json"
-
-	"github.com/creasty/defaults"
 	"github.com/go-playground/validator/v10"
-	"gopkg.in/yaml.v3"
+	"github.com/taimoorgit/moredefaults"
```
rocktavious, line 5: i would prefer we made this fork in the opslevel github org.  Also the long term goal being that we get the changes merged upstream.

gen/config/alertsource.json
rocktavious: ❤️  - i can get behind the centralized file as a middle ground.

utils.go
taimoorgit: @rocktavious 100%, let's transfer it over before merging.

property.go
taimoorgit: Yes! I forgot about this so it's an oversight on my end.

taimoorgit: Changed

utils.go
taimoorgit: Has been transferred: https://github.com/OpsLevel/moredefaults

gen/config/alertsource.json
taimoorgit: Changed code to do this in memory - no more files needed. Lets us cut down on the # of lines of code too, and writing file IO logic is annoying.

gen/parser.go
```diff
@@ -0,0 +1,104 @@
...
+func parse() error {
+	packages, err := parser.ParseDir(token.NewFileSet(), ".", nil, parser.ParseComments)
+	if err != nil {
+		return err
+	}
+	for _, file := range packages["opslevel"].Files {
+		ast.FileExports(file)
+		ast.Inspect(file, func(n ast.Node) bool {
+			var (
+				funcName    string
```
davidbloss, line 34: Can these `var` declarations be moved up out of the for loop? They are only be declared once either way and it's less to read through within the `for` loop.

gen/validate.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if _, ok := Combos[fn.Verb]; !ok {
+		complain("GetCase: unmapped verb on function '%s' with verb '%s'", fn.Name, fn.Verb)
+		return
+	} else {
+		combos = Combos[fn.Verb]
+	}
+	for _, combo := range combos {
+		var (
+			inOk, outOk = false, false
```
davidbloss, line 64:
```suggestion
			inOk, outOk bool
```
This will set both values to `false` since that's the "zero value" for bools

davidbloss: Also these var declarations can be moved to the top of `GetCase` with the others

taimoorgit: applied this suggestion, but in a new branch (no pr yet) where all the code gen stuff will stay.

rocktavious (approved):

PR #352, opened 2024-01-11
Db/update user input struct field

davidbloss:
## Changelog

Update type of `Role` field in `UserInput` struct. Was `*string`, needs to be `*UserRole` (which is an Enum).

- [X] List your changes here
- [ ] Make a `changie` entry, N/A generated code

### DIFF
--- a/gen.go
+++ b/gen.go
@@ -343,6 +343,7 @@ type {{.Name}} struct { {{range .InputFields }}
     {{- else if eq .Name "tags" }}TagInput
     {{- else if eq .Name "members" }}TeamMembershipUserInput
     {{- else if eq .Name "contacts" }}ContactInput
+    {{- else if eq .Type.Name "UserRole" }}UserRole
     {{- else if .Type.Name }}{{ template "converted_type" .Type }}
     {{- else }}{{ .Type.OfType.OfTypeName | convertPayloadType  }}{{ end -}} ` + "`" +
 		`json:"{{.Name | lowerFirst }}{{if ne .Type.Kind "NON_NULL"}},omitempty{{end}}"` +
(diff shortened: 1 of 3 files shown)


gen.go
```diff
@@ -369,6 +370,7 @@ type {{.Name}} struct { {{range .InputFields }}
       {{- else if or (eq "PredicateTypeEnum" .Type.Name) (hasSuffix "PredicateTypeEnum" .Type.OfType.OfTypeName) }}satisfies_jq_expression
       {{- else if or (eq "ServicePropertyTypeEnum" .Type.Name) (hasSuffix "ServicePropertyTypeEnum" .Type.OfType.OfTypeName) }}language
       {{- else if or (eq "UsersFilterEnum" .Type.Name) (hasSuffix "UsersFilterEnum" .Type.OfType.OfTypeName) }}last_sign_in_at
+      {{- else if or (eq "UserRole" .Type.Name) (hasSuffix "UserRole" .Type.OfType.OfTypeName) }}admin
```
rocktavious, line 373: oof. I foresee this being very brittle in the long term.  Do we have a plan to not have this?

davidbloss: Oh I agree, this is ripe for refactoring. I just wanted a usable (albeit ugly) first crack at this

taimoorgit: IMO this is fine, because if this ever does cause an unexpected change we should see it in the pull request diff. 

I would consider making this better outside the scope of this PR, since we are already using a lot of `hasSuffix` calls when generating enums.

taimoorgit (approved):

gen.go
rocktavious: As long as we have a plan to not have to remember to curate this.  This data "should" be reference-able from the schema.

I'm fine with it being a future fix because its not long term healthy to have this.

PR #356, opened 2024-01-21
Improve code quality/finish TODOs 

taimoorgit:
Applied most suggestions using GoLand. Rebased to main.

Non-code changes:

- Remove outdated TODOs
- Apply spelling, grammar, punctuation fixes, HTML/markdown mistakes
- GoLand suggestions
- Remove deprecated examples
- Remove `//#region` and `//#endregion` comments

In code:
- Stop using deprecated `graphql.Int` and replace it with `int`
- Fix inconsistent receiver names
- Remove unused functions, unused function arguments
- Remove functions already marked as deprecated


category.go
```diff
@@ -17,8 +17,8 @@ type CategoryConnection struct {
 	TotalCount int
 }
 
-func (self *Category) Alias() string {
-	return slug.Make(self.Name)
+func (cat *Category) Alias() string {
```
taimoorgit, line 20: 🦁

scalar.go
```diff
@@ -19,11 +19,11 @@ func NewID(id ...string) *ID {
 
 func (s ID) GetGraphQLType() string { return "ID" }
 
-func (s *ID) MarshalJSON() ([]byte, error) {
-	if *s == "" {
+func (s ID) MarshalJSON() ([]byte, error) {
```
taimoorgit, line 22: This was changed because it is considered a mistake to mix pointers and non-pointers in receiver functions. Our other `MarshalJSON` functions are not using pointers. 

> First, and most important, does the method need to modify the receiver? If it does, the receiver must be a pointer.
> Second is the consideration of efficiency. If the receiver is large, a big struct for instance, it will be much cheaper to use a pointer receiver.
> For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear.

https://go.dev/doc/faq#methods_on_values_or_pointers

---

**This change requires the test case for scalar_test.go to change.**

scalar_test.go
```diff
@@ -40,9 +40,9 @@ func TestMarshalID(t *testing.T) {
 	buf3, err3 := json.Marshal(case3)
 	// Assert
 	autopilot.Ok(t, err1)
-	autopilot.Equals(t, `{"key1":"","key3":null}`, string(buf1))
+	autopilot.Equals(t, `{"key1":null,"key3":null}`, string(buf1))
```
taimoorgit, line 43: These changes in the test expectation is justified. The test expectations before were incorrect.

```
type IDTester struct {
	Key1 ol.ID  `json:"key1"`
	Key2 ol.ID  `json:"key2,omitempty"`
	Key3 *ol.ID `json:"key3"`
	Key4 *ol.ID `json:"key4,omitempty"`
}

func TestMarshalID(t *testing.T) {
	// Arrange
	id1 := ol.NewID() <------ we're getting a pointer to a blank ID (string)
	id2 := ol.NewID("Z2lkOi8vMTIzNDU2Nzg5")
	case1 := IDTester{}
	case2 := IDTester{
		Key1: *id1, <------ this is now a concrete empty ID (""). key1 is not omit empty, so it should serialize to null.
		Key2: *id1,
		Key3: id1,
		Key4: id1,
	}
	case3 := IDTester{
		Key1: *id2, <------ empty ID (""). again, key 1 is not omit empty, so it should serialize to null.
		Key2: *id2,
		Key3: id2,
		Key4: id2,
	}
	// Act
	buf1, err1 := json.Marshal(case1)
	buf2, err2 := json.Marshal(case2)
	buf3, err3 := json.Marshal(case3)
	// Assert
	autopilot.Ok(t, err1)
	autopilot.Equals(t, `{"key1":null,"key3":null}`, string(buf1))
	autopilot.Ok(t, err2)
	autopilot.Equals(t, `{"key1":null,"key3":null,"key4":null}`, string(buf2))

```

Taskfile.yml
```diff
@@ -21,8 +21,8 @@ tasks:
     desc: Formatting and linting
     cmds:
       - test -z "$(gofumpt -d -e . | tee /dev/stderr)"
-      - go vet ./...
```
taimoorgit, line 24: removed this because go vet is already included as a step in `golangci-lint run` by default.

aliases.go
```diff
@@ -5,8 +5,6 @@ import (
 	"strings"
 )
 
-// #region Create
-// TODO: make sure duplicate aliases throw an error that we can catch
```
taimoorgit, line 9: Removed TODO because 1. we have a function call to remove duplicates and 2. it's not possible to make a "duplicate alias" anyways, so returning an error for that is unnecessary.

check.go
```diff
@@ -131,10 +129,6 @@ func NewCheckCreateInputTypeOf[T any](checkCreateInput CheckCreateInput) *T {
 	return newCheck
 }
 
-func (c *CheckCreateInput) GetCheckCreateInput() *CheckCreateInput {
```
taimoorgit, line 134: Removed this along with `GetCheckUpdateInput` because there are no use cases for both of these functions.

clientGQL.go
```diff
@@ -97,7 +91,6 @@ func (client *Client) Validate() error {
 		}
 	}
 	err := client.Query(&q, nil)
-	// TODO: we should probably use a custom OpsLevelClientError type - https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go
```
taimoorgit, line 100: Removed because we are fairing fine without this. If we did really want to do this, we should file a ticket since it's not a trivial thing to do that can be encompassed by a `// TODO` comment.

clientGQL_test.go
```diff
@@ -126,10 +126,6 @@ func (t *TestDataTemplater) ParseTemplatedString(contents string) string {
 	return strings.TrimSpace(data.String())
 }
 
-func IsValidJson(data string) bool {
```
taimoorgit, line 129: Removed - this is exposed in the test package only and is never called.

category.go
```diff
@@ -52,7 +46,7 @@ func (client *Client) GetCategory(id ID) (*Category, error) {
 	}
 	err := client.Query(&q, v, WithName("CategoryGet"))
 	if q.Account.Category.Id == "" {
-		err = fmt.Errorf("Category with ID '%s' not found!", id)
+		err = fmt.Errorf("category with ID '%s' not found", id)
```
taimoorgit, line 49: Starting errors with capital letters and ending with punctuation is discouraged in Go.

infra.go
```diff
@@ -92,15 +92,15 @@ func (i *InfrastructureResource) GetTags(client *Client, variables *PayloadVaria
...
+	return *NewID(infrastructureResource.Id)
 }
 
-func (i *InfrastructureResource) ResourceType() TaggableResource {
+func (infrastructureResource *InfrastructureResource) ResourceType() TaggableResource {
 	return TaggableResourceInfrastructureresource
 }
 
-func (client *Client) CreateInfrastructure(input InfraInput) (*InfrastructureResource, error) {
+func (client *Client) CreateInfrastructure(input InfraInput) (*InfrastructureResource, []OpsLevelWarnings, error) {
```
davidbloss, line 103: Adding warnings to the return params will break `createInfraCmd` in the cli and possibly elsewhere. We should discuss what to do with warnings and should more than likely capture these in logs since warnings don't generally effect the flow of a program

davidbloss:
This PR needs to be split into at least two PRs. One PR can have all the changes that don't actually change anything at all, like variable renaming and comment removals.

davidbloss (changes requested):
The added warnings return param should be removed, at least from this PR

infra.go
taimoorgit: Reverted this and re-added the `TODO` comment, I am avoiding adding logging since we don't want a library to log

check.go
davidbloss: This is used in the terraform provider and is the basis for `CheckCreateInputProvider` up above

taimoorgit:
@davidbloss addressed your feedback again re: the functions that got removed

service.go
```diff
@@ -555,22 +542,20 @@ func (client *Client) ListServicesWithProduct(product string, variables *Payload
 	return &q.Account.Services, nil
 }
 
-func NewTagArgs(tag string) TagArgs {
+func NewTagArgs(tag string) (TagArgs, error) {
```
davidbloss, line 545: This would cause [code in the Terraform provider](https://github.com/OpsLevel/terraform-provider-opslevel/blob/main/opslevel/datasource_opslevel_services.go#L52) to break. Also in 4 other parts of the same repo

service_test.go
```diff
@@ -808,7 +808,9 @@ func TestListServicesWithTag(t *testing.T) {
 
 	client := BestTestClient(t, "service/list_with_tag", requests...)
 	// Act
-	response, err := client.ListServicesWithTag(ol.NewTagArgs("app:worker"), nil)
+	tagArgs, err := ol.NewTagArgs("app:worker")
```
davidbloss, line 811: This should be reverted in line with the changes to the `NewTagArgs()` function signature

davidbloss (changes requested):
The added `error` return arg in the `NewTagArgs()` function signature needs to be removed, along with the related change in `service_test.go`. Other than that this looks good

service_test.go
taimoorgit: Will put this in a separate PR

service.go
taimoorgit: Will put this in a separate PR

taimoorgit:
> The added `error` return arg in the `NewTagArgs()` function signature needs to be removed, along with the related change in `service_test.go`. Other than that this looks good

@davidbloss just extracted this into https://github.com/OpsLevel/opslevel-go/pull/372 and removed it from here. This PR should be cosmetic-only now.

.changes/unreleased/Refactor-20240212-165401.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Refactor
```
davidbloss, line 1: Final note! This changie log should be removed then this is ready to merge

davidbloss (dismissed):
LGTM 🎉

davidbloss (approved):
LGTM! 🎉

PR #359, opened 2024-01-29
Remove log line from opslevel-go lib

taimoorgit:
This is the only `log.*` call in opslevel-go outside of tests and cache.go. 

An error wrap is more appropriate here and we should avoid logging from opslevel-go since it can affect other programs.


rocktavious (approved):
❤️  - does that mean we can remove zerolog as a dependency of opslevel-go?

taimoorgit:
> ❤️ - does that mean we can remove zerolog as a dependency of opslevel-go?

@rocktavious not quite because cache.go still uses it. 

Though IMO it would be better if cache.go returned `error`s instead of logging. That would help programs that call the cache be more aware of when the cache is having problems. I would like to make that change too, what do you think?

davidbloss (approved):

PR #365, opened 2024-02-05
JsonString - fix constructor case on objects/arrays wrapped in a string

taimoorgit:
## Issues

Part of: https://github.com/OpsLevel/kubectl-opslevel/pull/244 / https://github.com/OpsLevel/product/issues/8

## Changelog

- [ ] List your changes here
- [ ] Make a `changie` entry

### DIFF
--- a/json_test.go
+++ b/json_test.go
@@ -19,6 +19,39 @@ type JSONTester struct {
 	Key4 *ol.JSON `json:"key4,omitempty"`
 }
 
+func TestNewJsonString(t *testing.T) {
+	type TestCase struct {
+		Data   any
+		Output ol.JsonString
+	}
+	testCases := map[string]TestCase{
+		// object
+		"wrappedObject":      {`{"foo": "bar"}`, `{"foo": "bar"}`},
+		"wrappedObjectEmpty": {`{}`, `{}`},
+		// array
+		"wrappedArray":      {`["foo", "bar"]`, `["foo", "bar"]`},
+		"wrappedArrayEmpty": {`[]`, `[]`},
+		// string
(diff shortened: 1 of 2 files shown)


json_test.go
```diff
@@ -188,14 +246,3 @@ func TestUnmarshalJSONString(t *testing.T) {
 	autopilot.Equals(t, data5, result5.AsMap())
 	autopilot.Assert(t, err != nil, "The JSON string of type bool should be unable to unmarshalled into a float32")
 }
-
-func TestNewJSONSchema(t *testing.T) {
-	res, err := ol.NewJSONSchema(validStringContainingJSON)
-	resVal := *res
-	autopilot.Ok(t, err)
-	autopilot.Equals(t, resVal["name"], "Thomas")
```
taimoorgit, line 196: Remove because [all these `equals` checks are done by reflect.DeepEqual anyways](https://stackoverflow.com/a/18211675), which is what autopilot.Equals is doing

json_test.go
```diff
@@ -10,7 +10,34 @@ import (
 	"github.com/rocktavious/autopilot/v2023"
 )
 
-var validStringContainingJSON = `{"name":"Thomas","isIntern":false,"age":45,"access":{"aws":"admin","okta":"admin"},"tags":["org:engineering","team:platform"]}`
```
taimoorgit, line 13: Remove the `"age":45,` here because it is annoying to have `reflect.DeepEqual` fail to compare a int to a float64. This is handled by JSON marshal, we don't get any benefits from testing this on our own. We just need to make sure that keys and values are not being discarded/changed.

PR #368, opened 2024-02-10
Add suggestions enum and input from internal schema

taimoorgit:
## Changelog

- [ ] List your changes here
- [ ] Make a `changie` entry


rocktavious (changes requested):
👎  - we cannot use "internal" things in opslevel-go. We need to get the ruby teams to make the API public first and my understanding is they are not happy with the API as is so it needs significant work that is not on the short term roadmap

taimoorgit:
> 👎 - we cannot use "internal" things in opslevel-go. We need to get the ruby teams to make the API public first and my understanding is they are not happy with the API as is so it needs significant work that is not on the short term roadmap

Yep, this was just me experimenting. closing

PR #371, opened 2024-02-13
Fix JSON marshalling of an ID to be `null` (stop using unnecessary pointer)

taimoorgit:
## Issues

Extracted from https://github.com/OpsLevel/opslevel-go/pull/356

## TL;DR

The test expectation is wrong because we shouldn't be using a pointer receiver. 

## Explanation

These changes in the test expectation is justified. The test expectations before were incorrect.

```
type IDTester struct {
	Key1 ol.ID  `json:"key1"`
	Key2 ol.ID  `json:"key2,omitempty"`
	Key3 *ol.ID `json:"key3"`
	Key4 *ol.ID `json:"key4,omitempty"`
}

func TestMarshalID(t *testing.T) {
	// Arrange
	id1 := ol.NewID() <------ we're getting a pointer to a blank ID (string)
	id2 := ol.NewID("Z2lkOi8vMTIzNDU2Nzg5")
	case1 := IDTester{}
	case2 := IDTester{
		Key1: *id1, <------ this is now a concrete empty ID (""). key1 is not omit empty, so it should serialize to null.
		Key2: *id1,
		Key3: id1,
		Key4: id1,
	}
	case3 := IDTester{
		Key1: *id2, <------ empty ID (""). again, key 1 is not omit empty, so it should serialize to null.
		Key2: *id2,
		Key3: id2,
		Key4: id2,
	}
	// Act
	buf1, err1 := json.Marshal(case1)
	buf2, err2 := json.Marshal(case2)
	buf3, err3 := json.Marshal(case3)
	// Assert
	autopilot.Ok(t, err1)
	autopilot.Equals(t, `{"key1":null,"key3":null}`, string(buf1))
	autopilot.Ok(t, err2)
	autopilot.Equals(t, `{"key1":null,"key3":null,"key4":null}`, string(buf2))

```

## Changelog

- [x] Stop using unnecessary pointer
- [x] Fix test accordingly
- [x] Make a `changie` entry

### DIFF
--- a/scalar.go
+++ b/scalar.go
@@ -17,13 +17,13 @@ func NewID(id ...string) *ID {
 	return &output
 }
 
-func (s ID) GetGraphQLType() string { return "ID" }
+func (id ID) GetGraphQLType() string { return "ID" }
 
-func (s *ID) MarshalJSON() ([]byte, error) {
-	if *s == "" {
+func (id ID) MarshalJSON() ([]byte, error) {
+	if id == "" {
 		return []byte("null"), nil
 	}
-	return []byte(strconv.Quote(string(*s))), nil
+	return []byte(strconv.Quote(string(id))), nil
 }
 
 type Identifier struct {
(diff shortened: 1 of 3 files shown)


scalar.go
```diff
@@ -17,29 +17,29 @@ func NewID(id ...string) *ID {
...
 		return []byte("null"), nil
 	}
 	var out string
-	if i.Id != nil {
-		out = fmt.Sprintf(`{"id":"%s"}`, string(*i.Id))
+	if identifierInput.Id != nil {
+		out = fmt.Sprintf(`{"id":"%s"}`, string(*identifierInput.Id))
 	} else {
-		out = fmt.Sprintf(`{"alias":"%s"}`, string(*i.Alias))
+		out = fmt.Sprintf(`{"alias":"%s"}`, *identifierInput.Alias)
```
rocktavious, line 42: couldn't this cause a nil pointer deference panic?

scalar_test.go
```diff
@@ -40,9 +40,9 @@ func TestMarshalID(t *testing.T) {
 	buf3, err3 := json.Marshal(case3)
 	// Assert
 	autopilot.Ok(t, err1)
-	autopilot.Equals(t, `{"key1":"","key3":null}`, string(buf1))
+	autopilot.Equals(t, `{"key1":null,"key3":null}`, string(buf1))
```
rocktavious, line 43: null for IDs means "unset" in our API generally but what does `""` mean?  I'm concerned we are losing the ability to to create input types where `""` is the expected JSON.

What is the underlying impetus for this change?  golang has the ability to put functions on pointers or value types - theres no reason that I see restricting this.

PR #373, opened 2024-02-15
Db/gen remaining data structs

davidbloss:
## Changelog

- Update `gen.go` to generate all of the data structs 
- Add `interfaces.go` - these are GraphQL "interfaces" like `Check`
- Add `object.go` - these are GraphQL "objects", the majority of data structs

There are a lot of new structs being added that likely won't do anything for a while.
Some generated structs are replacing existing structs.
- Some of these are identical
- Some are the same with extra fields
- LOADS of testing updates, mostly GraphQL query updates
- There are 4 structs that don't have the `,omitempty` json tag (can be added) but wanted to vet those first

The 4 structs needing vetting are (all in `object.go`):
- `Level`
- `FilterPredicate`
- `ServiceDocument`
- `Tool`

It's a lot... 😁 Thank you ahead of time for review this! 🙏 
I also need to add changie logs

- [X] List your changes here
- [ ] Make a `changie` entry

## Tophatting

`task generate`


.golangci.yml
```diff
@@ -0,0 +1,3 @@
+linters:
```
davidbloss, line 1: Temporarily added because go vet points out an issue with `ServiceDocumentSource`.
Right now it looks like:
```go
# Embedded, both `IntegrationId` and `ServiceRepository` have a clashing `json:"id"` tag
type ServiceDocumentSource struct {
	IntegrationId     `graphql:"... on ApiDocIntegration"`
	ServiceRepository `graphql:"... on ServiceRepository"`
}
```
but it should look like
```go
type ServiceDocumentSource struct {
	IntegrationId     IntegrationId     `graphql:"... on ApiDocIntegration"`
	ServiceRepository ServiceRepository `graphql:"... on ServiceRepository"`
}
```

infra_test.go
```diff
@@ -52,14 +52,14 @@ func TestCreateInfra(t *testing.T) {
...
-	autopilot.Equals(t, string(id1), result.Id)
+	autopilot.Equals(t, id1, result.Id)
 	autopilot.Equals(t, "my-big-query", result.Name)
 }
 
 func TestGetInfra(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
-		`query InfrastructureResourceGet($all:Boolean!$input:IdentifierInput!){account{infrastructureResource(input: $input){id,aliases,name,type @include(if: $all),providerResourceType @include(if: $all),providerData @include(if: $all){accountName,externalUrl,providerName},owner @include(if: $all){... on Team{teamAlias:alias,id}},ownerLocked @include(if: $all),data @include(if: $all),rawData @include(if: $all)}}}`,
+		`query InfrastructureResourceGet($all:Boolean!$input:IdentifierInput!){account{infrastructureResource(input: $input){aliases,data,href,id,integration,lastSyncedAt,name,owner{... on Team{teamAlias:alias,id}},ownerLocked,providerData{accountName,externalUrl,providerName},providerResourceType,rawData,type}}}`,
```
rocktavious, line 62: This change is no good - the extra directive `@include(if: $all)` on things is needed because otherwise for a `list infra` query we'll be pulling back HUGE amounts of data

actions.go
```diff
@@ -100,7 +78,7 @@ func (client *Client) GetCustomAction(input string) (*CustomActionsExternalActio
 		"input": *NewIdentifier(input),
 	}
 	err := client.Query(&q, v, WithName("ExternalActionGet"))
-	if q.Account.Action.Id == "" {
+	if q.Account.Action.CustomActionsId.Id == "" {
```
rocktavious, line 81: This feels like a poor UX change.  Anyway we can make it like how it used to be?

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	CommitSha           string `graphql:"commitSha" json:"commitSha"`                     // The sha associated with the commit of the deploy.
+	CommittedAt         string `graphql:"committedAt" json:"committedAt"`                 // The time the commit happened.
+	CommitterEmail      string `graphql:"committerEmail" json:"committerEmail"`           // The email of the person who created the commit.
+	CommitterName       string `graphql:"committerName" json:"committerName"`             // The name of the person who created the commit.
+	DedupId             string `graphql:"dedupId" json:"dedupId"`                         // The deduplication ID provided to prevent duplicate deploys.
+	DeployNumber        string `graphql:"deployNumber" json:"deployNumber"`               // An identifier to keep track of the version of the deploy.
+	DeployUrl           string `graphql:"deployUrl" json:"deployUrl"`                     // The url the where the deployment can be found.
+	DeployedAt          string `graphql:"deployedAt" json:"deployedAt"`                   // The time the deployment happened.
```
rocktavious, line 293: this should be an ISO8601DateTime

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	DeployerId          string `graphql:"deployerId" json:"deployerId"`                   // An external id of who deployed.
+	DeployerName        string `graphql:"deployerName" json:"deployerName"`               // The name of who is responsible for the deployment.
+	Description         string `graphql:"description" json:"description"`                 // The given description of the deploy.
+	Environment         string `graphql:"environment" json:"environment"`                 // The environment in which the deployment happened in.
+	Id                  ID     `graphql:"id" json:"id"`                                   // The id of the deploy.
+	ProviderName        string `graphql:"providerName" json:"providerName"`               // The integration name of the deploy.
+	ProviderType        string `graphql:"providerType" json:"providerType"`               // The integration type used the deploy.
+	ProviderUrl         string `graphql:"providerUrl" json:"providerUrl"`                 // The url to the deploy integration.
+	Service             string `graphql:"service" json:"service"`                         // The service object the deploy is attached to.
```
rocktavious, line 303: this looks like its a `service` object so should be a `ServiceId`

document.go
```diff
@@ -1,16 +1,15 @@
 package opslevel
 
+// NOTE: test first, then replace this struct with commented out one below
```
taimoorgit, line 3: Is this going to be changed in this PR, or the next one?

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	CreatedAt   iso8601.Time `graphql:"createdAt" json:"createdAt"`     // The time this integration was created.
+	Id          ID           `graphql:"id" json:"id"`                   // The unique identifier for the integration.
+	InstalledAt iso8601.Time `graphql:"installedAt" json:"installedAt"` // The time that this integration was successfully installed, if null, this indicates the integration was not completed installed.
+	Name        string       `graphql:"name" json:"name"`               // The name of the integration.
+	Type        string       `graphql:"type" json:"type"`               // The type of the integration.
+}
+
+// HasDocumentationCheck represents .
+type HasDocumentationCheck struct {
+	Campaign        string       `graphql:"campaign" json:"campaign"`               // The campaign the check belongs to.
```
rocktavious, line 418: This is actually a `Campaign` object so it needs to be a `CampaignId` type

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	Id          ID           `graphql:"id" json:"id"`                   // The unique identifier for the integration.
+	InstalledAt iso8601.Time `graphql:"installedAt" json:"installedAt"` // The time that this integration was successfully installed, if null, this indicates the integration was not completed installed.
+	Name        string       `graphql:"name" json:"name"`               // The name of the integration.
+	Type        string       `graphql:"type" json:"type"`               // The type of the integration.
+}
+
+// Language represents a language that can be assigned to a repository.
+type Language struct {
+	Name  string  `graphql:"name" json:"name"`   // The name of the language.
+	Usage float32 `graphql:"usage" json:"usage"` // The percentage of the code written in that language.
```
rocktavious, line 513: do we know if Ruby is float64 or float32?  Honestly if it works float64 is better in golang because the standard math lib works with float64 and you have to bring in a specialized math32 external library to support math operations on float32 OR you have to convert / cast type all the time.

actions.go
davidbloss: Agreed

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	Filter                Filter       `graphql:"filter" json:"filter"`                               // The filter that the check belongs to.
+	Id                    ID           `graphql:"id" json:"id"`                                       // The id of the check.
+	Level                 Level        `graphql:"level" json:"level"`                                 // The level that the check belongs to.
+	Name                  string       `graphql:"name" json:"name"`                                   // The display name of the check.
+	Notes                 string       `graphql:"notes" json:"notes"`                                 // Additional information about the check.
+	Owner                 CheckOwner   `graphql:"owner" json:"owner"`                                 // The owner of the check.
+	Type                  CheckType    `graphql:"type" json:"type"`                                   // The type of check.
+	UpdateFrequency       string       `graphql:"updateFrequency" json:"updateFrequency"`             // The minimum frequency of the updates.
```
rocktavious, line 554: another one that is a specialized type

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	UpdateRequiresComment string       `graphql:"updateRequiresComment" json:"updateRequiresComment"` // Whether the check requires a comment or not.
```
rocktavious, line 555: this is suppose to be a boolean 🤔

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	Url                   string       `graphql:"url" json:"url"`                                     // The url to the check.
+}
+
+// ManualCheckFrequency represents .
+type ManualCheckFrequency struct {
+	FrequencyTimeScale FrequencyTimeScale `graphql:"frequencyTimeScale" json:"frequencyTimeScale"` // The time scale type for the frequency.
```
rocktavious, line 561: i don't see this field type in object.go - shouldn't it be here?

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	FrequencyValue     int                `graphql:"frequencyValue" json:"frequencyValue"`         // The value to be used together with the frequency scale.
+	StartingDate       iso8601.Time       `graphql:"startingDate" json:"startingDate"`             // The date that the check will start to evaluate.
+}
+
+// NewRelicIntegration represents .
+type NewRelicIntegration struct {
+	AccountKey                  string       `graphql:"accountKey" json:"accountKey"`                                   // The New Relic Account key for this integration.
+	AllowManualSyncAlertSources string       `graphql:"allowManualSyncAlertSources" json:"allowManualSyncAlertSources"` // Indicates if manual alert source synchronization can be triggered.
```
rocktavious, line 569: this one should also be a bool - this makes me a little scared about the generator is really not doing the right thing.  Can we figure out why?

object.go
```diff
@@ -0,0 +1,1131 @@
...
+	BaseUrl                     string       `graphql:"baseUrl" json:"baseUrl"`                                         // The API URL for New Relic API.
+	CreatedAt                   iso8601.Time `graphql:"createdAt" json:"createdAt"`                                     // The time this integration was created.
+	Id                          ID           `graphql:"id" json:"id"`                                                   // The unique identifier for the integration.
+	InstalledAt                 iso8601.Time `graphql:"installedAt" json:"installedAt"`                                 // The time that this integration was successfully installed, if null, this indicates the integration was not completed installed.
+	LastManualSyncAlertSources  string       `graphql:"lastManualSyncAlertSources" json:"lastManualSyncAlertSources"`   // The time that alert sources were last manually synchronized at.
```
rocktavious, line 574: another iso8601

PR #380, opened 2024-03-05
add examples for properties/property definitions

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/256

## Changelog

- [ ] make cli easier to use by adding example on property, property defn input 
- [ ] Make a `changie` entry


taimoorgit:
closing since this isn't practical to use in the CLI when it comes to property defs and assignments.

PR #383, opened 2024-03-09
Require cache to be instantiated with a logger

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/214

Improve code quality by removing the global variable and creating an instantiation function for cache.

Also allow library consumer to pass in a logger rather than being forced to use ours. 

## Changelog

- [x] Add `NewCacher()`, allows passing in a `zerolog.Logger`
- [x] Log errors in a structured field using `.Err(err)`
- [x] Make a `changie` entry

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -30,7 +30,6 @@ tasks:
       - gofumpt -w .
       - go mod tidy
       - golangci-lint run --fix
-      - nilaway -fix -test=false -include-pkgs='github.com/opslevel/opslevel-go' ./...
 
   setup:
     desc: Setup linter, formatter, etc. for local testing and CI
--- a/cache.go
+++ b/cache.go
@@ -3,10 +3,12 @@ package opslevel
 import (
 	"sync"
 
-	"github.com/rs/zerolog/log"
+	"github.com/rs/zerolog"
 )
 
+// Cacher is a cache mapping OpsLevel API objects to their alias. The API client never internally uses Cacher.
 type Cacher struct {
+	log          *zerolog.Logger
 	mutex        sync.Mutex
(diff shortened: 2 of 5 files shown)


Taskfile.yml
```diff
@@ -30,7 +30,6 @@ tasks:
       - gofumpt -w .
       - go mod tidy
       - golangci-lint run --fix
-      - nilaway -fix -test=false -include-pkgs='github.com/opslevel/opslevel-go' ./...
```
taimoorgit, line 33: Removing this step - nilaway still has false positives and I've never gotten the `-fix` to actually work since it's still an experimental linter. 

If something is ever triggered by nilaway, we should fix it manually since it's likely a larger issue.

cache.go
```diff
@@ -101,35 +120,35 @@ func (cacher *Cacher) TryGetInfrastructureSchema(alias string) (*InfrastructureR
 }
 
 func (cacher *Cacher) doCacheTiers(client *Client) {
-	log.Debug().Msg("Caching 'Tier' lookup table from API ...")
+	cacher.log.Debug().Msg("Caching 'Tier' lookup table from API ...")
 
 	data, dataErr := client.ListTiers()
 	if dataErr != nil {
-		log.Warn().Msgf("===> Failed to list all 'Tier' from API - REASON: %s", dataErr.Error())
```
taimoorgit, line 108: Changed these lines to use `.Err()` since it's convenient and makes the logs more structured.

Using `dataErr.Error()` and `%s` is not necessary, errors can be printed out using `%w`.

```
log.Warn().Msgf("===> Failed to list all 'Tier' from API - REASON: %w", dataErr)
```

PR #387, opened 2024-03-26
include infra Schema during infra update

davidbloss:
## Issues

The terraform provider is unable to update an infrastructure resource's `schema` field. This fixes that.

## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` - all tests pass

### DIFF
--- a/infra_test.go
+++ b/infra_test.go
@@ -146,6 +146,7 @@ func TestUpdateInfra(t *testing.T) {
 	autopilot.Equals(t, nil, err)
 	autopilot.Equals(t, string(id1), result.Id)
 	autopilot.Equals(t, "my-big-query", result.Name)
+	autopilot.Equals(t, "Database", result.Schema)
 }
 
 func TestDeleteInfra(t *testing.T) {
--- a/infra_test.go
+++ b/infra_test.go
@@ -127,7 +127,7 @@ func TestUpdateInfra(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation InfrastructureResourceUpdate($all:Boolean!$identifier:IdentifierInput!$input:InfrastructureResourceInput!){infrastructureResourceUpdate(infrastructureResource: $identifier, input: $input){infrastructureResource{id,aliases,name,type @include(if: $all),providerResourceType @include(if: $all),providerData @include(if: $all){accountName,externalUrl,providerName},owner @include(if: $all){... on Team{teamAlias:alias,id}},ownerLocked @include(if: $all),data @include(if: $all),rawData @include(if: $all)},warnings{message},errors{message,path}}}`,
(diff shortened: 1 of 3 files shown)


infra_test.go
```diff
@@ -127,7 +127,7 @@ func TestUpdateInfra(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation InfrastructureResourceUpdate($all:Boolean!$identifier:IdentifierInput!$input:InfrastructureResourceInput!){infrastructureResourceUpdate(infrastructureResource: $identifier, input: $input){infrastructureResource{id,aliases,name,type @include(if: $all),providerResourceType @include(if: $all),providerData @include(if: $all){accountName,externalUrl,providerName},owner @include(if: $all){... on Team{teamAlias:alias,id}},ownerLocked @include(if: $all),data @include(if: $all),rawData @include(if: $all)},warnings{message},errors{message,path}}}`,
-		`{"all": true, "identifier": { {{ template "id1" }}}, "input": { "ownerId": "{{ template "id1_string" }}", "data": "{\"endpoint\":\"https://google.com\",\"engine\":\"BigQuery\",\"name\":\"my-big-query\",\"replica\":false}" }}`,
+		`{"all": true, "identifier": { {{ template "id1" }}}, "input": { "ownerId": "{{ template "id1_string" }}", "schema": {"type": ""}, "data": "{\"endpoint\":\"https://google.com\",\"engine\":\"BigQuery\",\"name\":\"my-big-query\",\"replica\":false}" }}`,
```
taimoorgit, line 130: I noticed the create test uses a schema with a type, can that be added here?

https://github.com/OpsLevel/opslevel-go/blob/f4f74660988fc296978001baf6ea0b0c8c242e7a/infra_test.go#L25-L27

taimoorgit (changes requested):

taimoorgit (approved):
>The terraform provider is unable to update an infrastructure resource's schema field. This fixes that.

Change and test LGTM but do we even want to do this in the TF upgrade? Would this be out of scope?

davidbloss:
> > The terraform provider is unable to update an infrastructure resource's schema field. This fixes that.
> 
> Change and test LGTM but do we even want to do this in the TF upgrade? Would this be out of scope?

Why this is in scope:
- Looking at the [TF upgrade ticket](https://github.com/OpsLevel/team-platform/issues/308) acceptance criteria
   - "Data input to and returned from each resource should match the structure, types, etc..."
- Adding the missing `Schema` field here within the `UpdateInfrastructure()` func fixes a bug. When customers change their [opslevel_infrastructure](https://registry.terraform.io/providers/OpsLevel/opslevel/latest/docs/resources/infrastructure#schema) schema, specifically the `opslevel_infrastructure.schema` field it works on create but fails on update.
- The underlying schema hasn't changed (no input change) and this doesn't change the expected output from the terraform provider. It just fixes it.

PR #391, opened 2024-04-25
update listing services by filter, filter has no alias

davidbloss:
## Issues

Follow up to [this PR](https://github.com/OpsLevel/opslevel-go/pull/390)

## Changelog

Replaced InputIdentifier with id only. Filters do not have aliases

- [X] List your changes here
- [ ] Make a `changie` entry, N/A covered in previous PR

## Tophatting

`task test` - tests pass


service.go
```diff
@@ -379,7 +379,10 @@ func (client *Client) ListServices(variables *PayloadVariables) (*ServiceConnect
 	return &q.Account.Services, nil
 }
 
-func (client *Client) ListServicesWithFilter(filterIdentifier string, variables *PayloadVariables) (*ServiceConnection, error) {
+func (client *Client) ListServicesWithFilterId(filterId string, variables *PayloadVariables) (*ServiceConnection, error) {
```
rocktavious, line 382: lets not change the interface of the client.  We can kep the same function and just return the error if they don't present an ID with a message that the API currently doesn't support this.

rocktavious (changes requested):
lets limit the breaking change and just raise a validation error

rocktavious (approved):
LGTM

PR #395, opened 2024-05-02
replace float32 with float64 for OpsLevel terraform provider

davidbloss:
## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` - tests pass


.changes/unreleased/Refactor-20240502-084345.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Refactor
+body: update "Usage" field in Repository struct. Was float32, is now float64
```
rocktavious, line 2:
```suggestion
body: BREAKING - update "Usage" field in Repository struct. Was float32, is now float64
```

PR #397, opened 2024-05-03
introduce `NullableValue[T]`, allow unsetting fields on a service with `ServiceUpdateInputV2`

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/356

### Why?

We cannot unset tier alias, lifecycle alias in services because it can't tell the difference between those attributes being set to `null` or not included (since those fields are `*string` with `omitempty`).

This uses the same pattern as in `IdentifierInput` where if ID and alias are not set, it will JSON marshal into `null`.

## Changelog

- [x] Introduce `NullableValue[T]` (and constructors)
- [x] Deprecate `ServiceUpdateInput`, add `ServiceUpdateInputV2` (they are interchangeable)
- [x] Make a `changie` entry
- [x] Add table tests for service update, covering new and old update input struct types
- [x] Add table tests for `NullableValue[T]`
- [x] Tophatting test with terraform issue

### DIFF
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce NullableValue generic type (adds support for unsetting fields in update input structs)
+body: Introduce Nullable generic type allows string only (to support for unsetting fields in update input structs)
 time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce OptionalString for better object updates (support unsetting fields)
+body: Introduce NullableString for better object updates (support unsetting fields)
 time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce OptionalString for ServiceUpdateV2(), ServiceUpdateInputV2()
+body: Introduce OptionalString for better object updates
 time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: Introduce OptionalString for NewServiceUpdate(), NewServiceUpdateInput()
+time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce NullableString for better object updates (support unsetting fields)
+body: Introduce NullableValue generic type (adds support for unsetting fields in update input structs)
 time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce OptionalString for better object updates
+body: Introduce OptionalString for better object updates (support unsetting fields)
 time: 2024-05-03T13:13:11.992999-04:00
--- a/.changes/unreleased/Feature-20240503-131311.yaml
+++ b/.changes/unreleased/Feature-20240503-131311.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: Introduce OptionalString for NewServiceUpdate(), NewServiceUpdateInput()
+body: Introduce OptionalString for ServiceUpdateV2(), ServiceUpdateInputV2()
 time: 2024-05-03T13:13:11.992999-04:00
--- a/scalar.go
+++ b/scalar.go
@@ -75,25 +75,25 @@ func IsID(value string) bool {
 	return strings.HasPrefix(string(decoded), "gid://")
 }
 
-// OptionalString is implemented using a bool indicating whether the field should be unset, this is required for
+// NullableString is implemented using a bool indicating whether the field should be unset, this is required for
 // backwards compatability so that we can still differentiate between "" and nil
-type OptionalString struct {
+type NullableString struct {
 	Value   string
 	SetNull bool
 }
 
-// NewOptionalString will create an optional string provided an input string, if no input string is provided the result
+// NewNullableString will create an optional string provided an input string, if no input string is provided the result
 // will json marshal into `null`
-func NewOptionalString(input ...string) *OptionalString {
+func NewNullableString(input ...string) *NullableString {
 	if len(input) == 1 {
-		return &OptionalString{Value: input[0]}
+		return &NullableString{Value: input[0]}
 	}
-	return &OptionalString{SetNull: true}
+	return &NullableString{SetNull: true}
 }
 
-func (optionalString *OptionalString) MarshalJSON() ([]byte, error) {
-	if optionalString.SetNull {
+func (nullableString *NullableString) MarshalJSON() ([]byte, error) {
+	if nullableString.SetNull {
 		return []byte("null"), nil
 	}
-	return []byte(strconv.Quote(optionalString.Value)), nil
+	return []byte(strconv.Quote(nullableString.Value)), nil
 }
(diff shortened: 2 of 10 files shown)


service_test.go
```diff
@@ -283,6 +283,26 @@ func TestUpdateService(t *testing.T) {
 	autopilot.Equals(t, "Foo", result.Name)
 }
 
+func TestNewUpdateService(t *testing.T) {
+	// Arrange
+	testRequest := autopilot.NewTestRequest(
+		`mutation ServiceUpdate($input:NewServiceUpdateInput!){serviceUpdate(input: $input){service{apiDocumentPath,description,framework,htmlUrl,id,aliases,language,lifecycle{alias,description,id,index,name},managedAliases,name,owner{alias,id},parent{id,aliases},preferredApiDocument{id,htmlUrl,source{... on ApiDocIntegration{id,name,type},... on ServiceRepository{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},timestamps{createdAt,updatedAt}},preferredApiDocumentSource,product,repos{edges{node{id,defaultAlias},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},{{ template "pagination_request" }},totalCount},tags{nodes{id,key,value},{{ template "pagination_request" }},totalCount},tier{alias,description,id,index,name},timestamps{createdAt,updatedAt},tools{nodes{category,categoryAlias,displayName,environment,id,url,service{id,aliases}},{{ template "pagination_request" }},totalCount}},errors{message,path}}}`,
```
rocktavious, line 289: `$input:NewServiceUpdateInput!` <- will not actually work here our API does not have a type `NewServiceUpdateInput`

service_test.go
```diff
@@ -304,6 +324,31 @@ func TestUpdateServiceWithSystem(t *testing.T) {
 	autopilot.Equals(t, "Foo", result.Name)
 }
 
+func TestNewUpdateServiceWithFields(t *testing.T) {
+	// Arrange
+	testRequest := autopilot.NewTestRequest(
+		`mutation ServiceUpdate($input:ServiceUpdateInputV2!){serviceUpdate(input: $input){service{apiDocumentPath,description,framework,htmlUrl,id,aliases,language,lifecycle{alias,description,id,index,name},managedAliases,name,owner{alias,id},parent{id,aliases},preferredApiDocument{id,htmlUrl,source{... on ApiDocIntegration{id,name,type},... on ServiceRepository{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},timestamps{createdAt,updatedAt}},preferredApiDocumentSource,product,repos{edges{node{id,defaultAlias},serviceRepositories{baseDirectory,displayName,id,repository{id,defaultAlias},service{id,aliases}}},{{ template "pagination_request" }},totalCount},tags{nodes{id,key,value},{{ template "pagination_request" }},totalCount},tier{alias,description,id,index,name},timestamps{createdAt,updatedAt},tools{nodes{category,categoryAlias,displayName,environment,id,url,service{id,aliases}},{{ template "pagination_request" }},totalCount}},errors{message,path}}}`,
```
rocktavious, line 330: No `ServiceUpdateInputV2` will also not work because thats not a type in our graphql API.  There is a way to control what is generated here i just don't recall of the top of my head its over in the framework docs.  We can ONLY use `ServiceUpdateInput` here.  But like i said there is a way to override this generation for the go type named `ServiceUpdateInputV2`

.changes/unreleased/Feature-20240503-131311.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Feature
+body: Introduce OptionalString for better object updates (support unsetting fields)
```
rocktavious, line 2:
```suggestion
body: Introduce OptionalString to support being able to setting a field to the JSON value 'null'
```

scalar.go
```diff
@@ -74,3 +74,26 @@ func IsID(value string) bool {
 	}
 	return strings.HasPrefix(string(decoded), "gid://")
 }
+
+// OptionalString is implemented using a bool indicating whether the field should be unset, this is required for
+// backwards compatability so that we can still differentiate between "" and nil
+type OptionalString struct {
```
rocktavious, line 80: What if we call this `NullableString` ?

service.go
```diff
@@ -663,13 +663,14 @@ func (client *Client) ListServicesWithTier(tier string, variables *PayloadVariab
 	return &q.Account.Services, nil
 }
 
-func (client *Client) UpdateService(input ServiceUpdateInput) (*Service, error) {
+func (client *Client) UpdateService(input ServiceUpdater) (*Service, error) {
```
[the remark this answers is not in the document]
taimoorgit: There's a disadvantage to using `any` which is that it isn't obvious what would be accepted here as input without looking at the source code, and it would also ruin autocomplete. We have an unnecessary `default` case here that can be avoided by using the `ServiceUpdater` interface. The random functions can easily be auto-generated and are public so they will not be visible to end users.

scalar.go
```diff
@@ -74,3 +75,29 @@ func IsID(value string) bool {
 	}
 	return strings.HasPrefix(string(decoded), "gid://")
 }
+
+// NullableValue can be used to unset a value using an OpsLevel input struct type, should always be instantiated using a constructor.
+type NullableValue[T any] struct {
+	Value   T
+	SetNull bool
+}
```
rocktavious, line 83: Is there a reason we are not doing

```suggestion
type NullableValue[T any] struct {
	sql.Null[T]
}
```

scalar_test.go
```diff
@@ -221,3 +224,146 @@ func TestNewIdentifierArray(t *testing.T) {
 	autopilot.Equals(t, "my-service", *result[0].Alias)
 	autopilot.Equals(t, ol.ID("Z2lkOi8vMTIzNDU2Nzg5"), *result[1].Id)
 }
+
+func TestNullValueString(t *testing.T) {
```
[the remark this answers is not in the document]
rocktavious: Optionally since we know we only want `Nullable` for string - you could add a type constraint if you are worried about people trying to use this for other types.  That way its future flexible to expand out to other types when we are ready.

scalar.go
```diff
@@ -74,3 +75,29 @@ func IsID(value string) bool {
...
+		SetNull: true,
+	}
+}
+
+func NewNullableValue[T any](value T) *NullableValue[T] {
+	return &NullableValue[T]{
+		Value:   value,
+		SetNull: false,
+	}
+}
```
rocktavious, line 103: You could potentially make this a global var is you rearrange the constructors like `sql` does

```suggestion
var Null = NewNullable(nil, true)

func NewNullable[T any](value T, isNull bool) *NullableValue[T] {
	return &NullableValue[T]{
	    Value: value
		SetNull: isNull,
	}
}

func NewNullableFrom[T any](value T) *NullableValue[T] {
	return &NullableValue[T]{
		Value:   value,
		SetNull: false,
	}
}
```

Then if you ever need to use just `Null` you can

```
input := opslevel.ServiceUpdateInput2{
    Description: opslevel.Null,
    Framework: opslevel.NewNullableFrom[string]("Cobra")
}
```

service_test.go
```diff
@@ -264,44 +265,94 @@ func TestCreateServiceWithParentSystem(t *testing.T) {
...
+				Name:           ol.NewNullableValue[string]("Hello World"),
+				Description:    ol.NewNullableValue("The quick brown fox"),
+				Framework:      ol.NewNullableValue("django"),
+				TierAlias:      ol.NewNullableValue("tier_4"),
+				LifecycleAlias: ol.NewNullableValue("pre-alpha"),
+			},
+		},
 
-	client := BestTestClient(t, "service/update_with_system", testRequest)
+		// unsetting fields does not work properly on old update input...
```
rocktavious, line 305: comment is better placed in the test "name"

rocktavious (approved):
Overall its good.  A few nit picks on naming and constructor organization but if you feel strongly i'm ok with it.

scalar_test.go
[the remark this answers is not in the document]
taimoorgit: @rocktavious pushed a commit to add the type constraint so that only `string` is supported for now. Like you said we can expand the constraint and add the extra tests later.

scalar.go
```diff
@@ -74,3 +75,29 @@ func IsID(value string) bool {
 	}
 	return strings.HasPrefix(string(decoded), "gid://")
 }
+
+// NullableValue can be used to unset a value using an OpsLevel input struct type, should always be instantiated using a constructor.
+type NullableValue[T any] struct {
```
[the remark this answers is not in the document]
taimoorgit: Renamed this to `Nullable`, rename construtors to be more clear.

`NewNullvalue` -> `NewNull` (emphasizes that the returned is to be treated as null.)

`NewNullableValue` -> `NewNullableWithValue` (emphasizes that the returned contains a value.)

scalar.go
taimoorgit: Renamed and added comment for the constructors

rocktavious (approved):

PR #398, opened 2024-05-05
add Alias type to match GraphQL scalar type and its behavior

davidbloss:
## Issues

[fix unsetting service lifecycleAlias and tierAlias fields](https://github.com/OpsLevel/team-platform/issues/356)

## Changelog

Add `String` type to match GraphQL scalar - just like we do with the `ID` scalar defined in our GraphQL schema.
`String` can replace all `string` instances if we want.

This change is in service of [this terraform provider PR](https://github.com/OpsLevel/terraform-provider-opslevel/pull/330).

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` - tests pass


davidbloss:
This solution is too narrow in scope for the repo's long term needs

PR #406, opened 2024-05-17
make Predicate in repo_search a pointer like other checks

davidbloss:
## Issues

This is the only check that had a non-pointer `Predicate` in a Fragment. This caused issues in the Terraform provider.

## Changelog

- [ ] List your changes here
- [ ] Make a `changie` entry

## Tophatting

`task test`


rocktavious (approved):
LGTM - can you add a changie entry about this

PR #409, opened 2024-05-31
add Validate func to Predicate struct

davidbloss:
## Changelog

Add `Validate()` to `Predicate` struct. Useful for Terraform provider config validation

- [X] List your changes here
- [X] Make a `changie` entry


predicates.go
taimoorgit, line 1: `predicate.go` (no s) would be a more idiomatic name

taimoorgit (approved):

PR #415, opened 2024-06-10
enable unsetting predicates of type PredicateUpdateInput with custom …

davidbloss:
## Issues

Change needed for [Terraform provider fix](https://github.com/OpsLevel/terraform-provider-opslevel/pull/367)

## Changelog

Add `MarshalJSON()` to `PredicateUpdateInput` to enable unsetting predicates during update operations.

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test` - tests pass

### DIFF
--- a/check_test.go
+++ b/check_test.go
@@ -583,14 +583,13 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
 		"UpdateToolUsageNullPredicates": {
 			fixture: autopilot.NewTestRequest(
 				`mutation CheckToolUsageUpdate($input:CheckToolUsageUpdateInput!){checkToolUsageUpdate(input: $input){check{category{id,name},description,enableOn,enabled,filter{id,name,connective,htmlUrl,predicates{key,keyData,type,value,caseSensitive}},id,level{alias,description,id,index,name},name,notes: rawNotes,owner{... on Team{alias,id}},type,... on AlertSourceUsageCheck{alertSourceNamePredicate{type,value},alertSourceType},... on CustomEventCheck{integration{id,name,type},passPending,resultMessage,serviceSelector,successCondition},... on HasRecentDeployCheck{days},... on ManualCheck{updateFrequency{frequencyTimeScale,frequencyValue,startingDate},updateRequiresComment},... on RepositoryFileCheck{directorySearch,filePaths,fileContentsPredicate{type,value},useAbsoluteRoot},... on RepositoryGrepCheck{directorySearch,filePaths,fileContentsPredicate{type,value}},... on RepositorySearchCheck{fileExtensions,fileContentsPredicate{type,value}},... on ServiceOwnershipCheck{requireContactMethod,contactMethod,tagKey,tagPredicate{type,value}},... on ServicePropertyCheck{serviceProperty,propertyValuePredicate{type,value}},... on TagDefinedCheck{tagKey,tagPredicate{type,value}},... on ToolUsageCheck{toolCategory,toolNamePredicate{type,value},toolUrlPredicate{type,value},environmentPredicate{type,value}},... on HasDocumentationCheck{documentType,documentSubtype}},errors{message,path}}}`,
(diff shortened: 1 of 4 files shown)


predicate.go
```diff
@@ -26,3 +26,16 @@ func (p *Predicate) Validate() error {
 	}
 	return nil
 }
+
+func (p *PredicateUpdateInput) MarshalJSON() ([]byte, error) {
```
rocktavious, line 30: can we get a test case that shows this works as expected?

predicate.go
```diff
@@ -26,3 +26,16 @@ func (p *Predicate) Validate() error {
...
+	if p == nil || p.Type == nil || *p.Type == "" {
+		return []byte("null"), nil
+	}
+	var predicateAsJson string
```
taimoorgit, line 34: I recommend a different approach,

Make a `map[string]string` from the struct and then just return the output of `json.Marshal` on that.

That way the `fmt.Sprintf` statements will not be necessary. Also, since JSON marshal usually returns JSON without any spacing I recommend that we do that too.

davidbloss: Good call on the spacing. I've added that.

Calling `json.Marshal()` from within `json.Marshal()` is recursive and causes a crash. Where would `json.Marshal()` be called in this other approach?

predicate_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	ol "github.com/opslevel/opslevel-go/v2024"
+	"github.com/rocktavious/autopilot/v2023"
+)
+
+func TestJsonMarshalPredicateUpdateInputNull(t *testing.T) {
```
rocktavious, line 11: Ahh this isn't want I meant but is a good test to have too.  What I wanted to see was the actual marshaling of check input as done by the graphql framework.  Like a test on the Check type that uses predicates.

predicate.go
taimoorgit: The code would look something like,

```
func (p *PredicateUpdateInput) MarshalJSON() ([]byte, error) {
  m := map[string]string{
    "type": *p.Type,
  }

  if p.Value != nil {
    m["value"]: *p.Value,
  }

  return JSON.Marshal(m)
}
```

There shouldn't be any recursion since we are marshalling the `map[string]string` and not the `PredicateUpdateInput`.

check_test.go
```diff
@@ -580,6 +580,22 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
...
+				`{ "input": { "id": "Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpIYXNPd25lci8yNDE4", {{ template "check_base_vars" }}, "toolCategory": "metrics", "toolNamePredicate": null, "toolUrlPredicate": null, "environmentPredicate": null } }`,
+				`{ "data": { "checkToolUsageUpdate": { "check": { "category": { "id": "Z2lkOi8vb3BzbGV2ZWwvQ2F0ZWdvcnkvNjA1", "name": "Performance" }, "description": "The service is using 'datadog' as a metrics tool in the 'production' environment.", "enabled": true, "id": "Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpIYXNPd25lci8yNDE4", "level": { "alias": "bronze", "description": "Services in this level satisfy critical checks. This is the minimum standard to ship to production.", "id": "Z2lkOi8vb3BzbGV2ZWwvTGV2ZWwvMzE3", "index": 1, "name": "Bronze" }, "name": "Hello World" }, "errors": [] } } }`,
+			),
+			endpoint: "check/update_tool_usage_null_predicates",
+			body: func(c *ol.Client) (*ol.Check, error) {
+				checkToolUsageUpdateInput := ol.NewCheckUpdateInputTypeOf[ol.CheckToolUsageUpdateInput](checkUpdateInput)
+				checkToolUsageUpdateInput.ToolCategory = ol.RefOf(ol.ToolCategoryMetrics)
+				checkToolUsageUpdateInput.ToolNamePredicate = &ol.PredicateUpdateInput{}
```
rocktavious, line 593: Awesome this is good - can you remove one of these so ensure "when not set" the field is omitted from the variables.

rocktavious (approved):

PR #419, opened 2024-06-14
Db/update filter predicate validation

davidbloss:
## Changelog

For filter predicates of type `greater_than_or_equal_to` and `less_than_or_equal_to`, verify the predicate value is a number

- [X] List your changes here
- [ ] Make a `changie` entry

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -43,9 +43,7 @@ func (filterPredicate *FilterPredicate) Validate() error {
 		PredicateTypeEnumEndsWith,
 		PredicateTypeEnumStartsWith,
 	}
-	if filterPredicate.CaseSensitive != nil &&
-		*filterPredicate.CaseSensitive &&
-		!slices.Contains(caseSensitiveTypes, filterPredicate.Type) {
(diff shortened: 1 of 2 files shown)


filters.go
```diff
@@ -35,9 +36,22 @@ func (filterPredicate *FilterPredicate) Validate() error {
 		PredicateTypeEnumEndsWith,
 		PredicateTypeEnumStartsWith,
 	}
-	if filterPredicate.CaseSensitive != nil && !slices.Contains(caseSensitiveTypes, filterPredicate.Type) {
+	if filterPredicate.CaseSensitive != nil &&
```
taimoorgit, line 39: Why is `*filterPredicate.CaseSensitive` necessary? I think the predicate would still be incorrect if if case sensitive was set to false.

taimoorgit (changes requested):

filters.go
davidbloss: This is true. I'm testing out some Terraform workflows and this was a workaround that I don't think we need

taimoorgit (approved):

PR #422, opened 2024-06-17
add generic ReconcileTags member function to client

davidbloss:
## Issues

[combine tags funcs to reconcileXTags](https://github.com/OpsLevel/team-platform/issues/363)

## Changelog

Add generic `ReconcileTags()` function that can manage tags on any resource that implements `TaggableResourceInterface`

- [X] List your changes here
- [X] Make a `changie` entry


tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+		tagConnection, err = resource.GetTags(client, nil)
+	case *Service:
+		tagConnection, err = resource.GetTags(client, nil)
+	case *System:
+		tagConnection, err = resource.GetTags(client, nil)
+	case *Team:
+		tagConnection, err = resource.GetTags(client, nil)
+	case *User:
+		tagConnection, err = resource.GetTags(client, nil)
+	default:
```
taimoorgit, line 234: I don't think that this `switch` is necessary since the `resourceType` is guaranteed to be of the `TaggableResourceInterface` (therefore the default case will never be hit).

tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+		return assignedTags, fmt.Errorf("not a taggable resource type: %s" + string(resourceType.ResourceType()))
+	}
+	if err != nil {
+		return assignedTags, err
+	}
+	if tagConnection == nil {
+		return assignedTags, fmt.Errorf("no tags found on %s with id '%s'", string(resourceType.ResourceType()), resourceType.ResourceId())
```
taimoorgit, line 241: Should this return an error? IMO just returning an empty array is fine.

tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+	}
+
+	// delete tags found in resource but not listed in tagsWanted
+	for _, tag := range tagConnection.Nodes {
+		if !slices.Contains(tagsWanted, tag.Flatten()) {
+			if err := client.DeleteTag(tag.Id); err != nil {
+				return assignedTags, err
```
taimoorgit, line 248: Since this function performs multiple API calls if one delete call fails I don't think the entire function should stop.

The errors join can be used here (link to example) - https://pkg.go.dev/errors#Join

tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+			}
+		}
+	}
+
+	// format tags listed in Terraform config but not found in service
+	tagInput := map[string]string{}
+	for _, tag := range tagsWanted {
+		parts := strings.Split(tag, ":")
+		if len(parts) != 2 {
+			return assignedTags, fmt.Errorf("[%s] invalid tag, should be in format 'key:value' (only a single colon between the key and value, no spaces or special characters)", tag)
```
taimoorgit, line 258: 1. I think it's odd that we check this on tag insertion but not tag deletion.
2. Since we always want key value pairs, we can just make it so that the function argument `tagsWanted` is already a map of key value pairs? No need to split and parse the string that way and we have a solid schema for what the input should look like.

tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+		}
+		key := parts[0]
+		value := parts[1]
+		tagInput[key] = value
+	}
+	// assign tags listed in Terraform config but not found in service
+	assignedTags, err = client.AssignTags(string(resourceType.ResourceId()), tagInput)
+	if err != nil {
+		return assignedTags, err
```
taimoorgit, line 267: Same comment about error joining

taimoorgit (changes requested):
Have questions/suggestions about the design of the function

tags.go
taimoorgit: Also it's unclear to me if the returned `tagConnection` could be nil (not the fault of this PR)

tags.go
davidbloss: Shouldn't the expectation here be that a delete tag call actually deletes each tag? What is the goal of bypassing errors to accumulate further errors?

tags.go
davidbloss: Here we expect `tagConnection` to be a `TagConnection` with a `Nodes` field of type `[]Tag`. This is an extra check that would most likely be covered by the `err` check above. An empty `tagConnection.Nodes` slice is fine but, but `tagConnection` should not be nil. This also gets ahead of a nil pointer dereference in the for loop below

tags.go
davidbloss: I agree that this would probably never happen. This `default` case was added mostly as extra guardrails for really odd edge cases

tags.go
davidbloss: 1. Tag insertion takes a slice of tags where each tag is a single string of format `key:value`. Tag deletion gets its tags from an API call where each `tag.Id` is known. These two logic streams take two different types of inputs and have different expectations.
2. We _could_ change tags from strings to a mapping. But that expands the scope of this task to require changes from all the calling functions. We could also build up the `Tag` struct with constructors that handle this.

tags.go
davidbloss: There's a single err here. What needs joining?

tags.go
rocktavious: I think the case Taimoor is going after is this:

Say you have 5 tags to delete
1.) Success
2.) Failure <- in your code it stops here but if it didn't
3.) Success
4.) Failure
5.) Success <- If we accumulate the errors we can report the end user that 2 tags were unable to be deleted for N reasons.

tags.go
rocktavious: @taimoorgit `(therefore the default case will never be hit)` is a VERY bad assumption to make.  In the future if we add a new type that flows through this codepath without the default case we'd could get unexpected behaviour or worse a crash.  You should ALWAYS handle the default case when you make a `switch` no matter if you think it will never be hit because you don't know what the future holds.

tags.go
taimoorgit: You are right, `AssignTags` only returns one error since it's just a single API call - resolving this

tags.go
taimoorgit: Yes exactly like what we do here in `client.CreateTags()` - https://github.com/OpsLevel/opslevel-go/blob/8016ef2dfc61c16bd04913d73eb651794e4d4d1e/tags.go#L149

tags.go
rocktavious: Now that being side - @davidbloss since you are taking in a interface and the expected func on that interface is `GetTags` then yes this switch statement is wholly unnecessary.  Thats the purpose of having the function signature take in an interface.

tags.go
taimoorgit: Since we are only using `range` [which is nil-safe it will not result in any nil pointer exceptions](https://docs.datadoghq.com/code_analysis/static_analysis_rules/go-best-practices/avoid-nil-check-loop/#description)

tags.go
```diff
@@ -200,3 +206,66 @@ func (client *Client) DeleteTag(id ID) error {
...
+		tagConnection, err = resource.GetTags(client, nil)
+	case *Team:
+		tagConnection, err = resource.GetTags(client, nil)
+	case *User:
+		tagConnection, err = resource.GetTags(client, nil)
+	default:
+		return assignedTags, fmt.Errorf("not a taggable resource type: %s" + string(resourceType.ResourceType()))
+	}
+	if err != nil {
+		return assignedTags, err
```
taimoorgit, line 220: [non-blocking] I suggest wrapping these errors since there are different API client calls that cause them

tags.go
taimoorgit: ^This is incorrect since we are accessing `nil.something`

tags.go
taimoorgit: Since `ReconcileAliases(0` also exits on the first error and returns the first error we can ignore this for now.

tags.go
taimoorgit: 1. Understood

2. I still think that forcing them to use key value pairs in a `map` is the way to go but they will either have to validate the input before calling this function or after calling the function so validating it in the function still works.

aliases.go
```diff
@@ -78,3 +104,57 @@ func (client *Client) DeleteAlias(input AliasDeleteInput) error {
...
+
+// ReconcileAliases manages aliases API operations for AliasOwnerInterface implementations
+//
+// Aliases not in 'aliasesWanted' will be deleted, new tags from 'aliasesWanted' will be created. Reconciled aliases are returned.
+func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliasesWanted []string) ([]string, error) {
+	var deleteAliasFunc func(string) error
+	var existingAliases []string
+	var resourceId ID
+
+	switch resource := resourceType.(type) {
```
taimoorgit, line 116: Can remove this switch statement just like the one in `ReconcileTags()` by 

1. Making the `AliasOwnerInterface` provide a `ResourceID()` function (returns an `ID` always so we don't have the case for InfrastructureResource)
2. Add an `Aliases()` function so we don't have to use this switch statement to get `ManagedAliases` vs `Aliases`

Makes for a better opslevel-go experience so that customers don't also have to write these types of `switch` statements.

davidbloss: @taimoorgit Customers don't have to write these kinds of switch statements. This is a very specific case that's being addressed and there is no apparent need for an interface at all. These added functions add no apparent value outside this function

aliases.go
```diff
@@ -2,9 +2,14 @@ package opslevel
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 )
 
+type AliasOwnerInterface interface {
+	AliasOwnerType() AliasOwnerTypeEnum
```
rocktavious, line 10: I don't see this function being used anywhere.  Am i missing something?

aliases.go
```diff
@@ -44,20 +49,41 @@ func (client *Client) CreateAlias(input AliasCreateInput) ([]string, error) {
 	return output, HandleErrors(err, m.Payload.Errors)
 }
 
+func (client *Client) DeleteDomainAlias(alias string) error {
```
rocktavious, line 52: This seems like a different paradigm then what was built for `TaggableResource` where you pass in the client to a function on the interface.  Why 2 paradigms?

aliases.go
rocktavious: This feels really awkward.  It feels optimized for having a single function on the client `ReconcileAliases` when the purpose of an interface is to have a single function on an object that you call to do the work.

I wonder if it would feel better and also be easier to use if the interface was placing the `ReconcileAliases` interface on the individual resource object and then its implemented per resource object (just like a standard interface would be) something the end up being used like this

```
domain, _ := client.GetDomain("my-domain")
domain.ReconcileAliases(client, []string{"one","three","four"})
```

aliases.go
```diff
@@ -78,3 +104,57 @@ func (client *Client) DeleteAlias(input AliasDeleteInput) error {
...
+			}
+		}
+	}
+
+	newServiceAliases := []string{}
+	for _, aliasWanted := range aliasesWanted {
+		if !slices.Contains(existingAliases, aliasWanted) {
+			newServiceAliases = append(newServiceAliases, aliasWanted)
+		}
+	}
```
rocktavious, line 157: Then you can extract out this functionality to a more simple function that takes in `currentAliases`, `desiredAliases` and return `aliasesToCreate` and `aliasesToDelete`

So you do something like
```
aliasesToCreate, aliasesToDelete := reconcileAliases(domain.Aliases, []string{"one","three","five"})
```

This also makes it so that `reconcileAliases` is easily unit testable with different string slices as input to ensure the reconcile logic is correct.

tags.go
```diff
@@ -200,3 +206,48 @@ func (client *Client) DeleteTag(id ID) error {
 	err := client.Mutate(&m, v, WithName("TagDelete"))
 	return HandleErrors(err, m.Payload.Errors)
 }
+
+// ReconcileTags manages tags API operations for TaggableResourceInterface implementations
+//
+// Tags not in 'tagsWanted' will be deleted, new tags from 'tagsWanted' will be created. Reconciled tags are returned.
+func (client *Client) ReconcileTags(resourceType TaggableResourceInterface, tagsWanted []string) ([]Tag, error) {
```
rocktavious, line 213: Why does this take in a `[]string` for the tags instead of `[]Tag` ?

tags.go
```diff
@@ -200,3 +206,48 @@ func (client *Client) DeleteTag(id ID) error {
...
+		if !slices.Contains(tagsWanted, tag.Flatten()) {
+			if err := client.DeleteTag(tag.Id); err != nil {
+				return assignedTags, err
+			}
+		}
+	}
+
+	// format tags listed in Terraform config but not found in service
+	tagInput := map[string]string{}
+	for _, tag := range tagsWanted {
```
rocktavious, line 237: if we took in `[]Tag` this would go away completely.

tags.go
```diff
@@ -200,3 +206,48 @@ func (client *Client) DeleteTag(id ID) error {
...
+		parts := strings.Split(tag, ":")
+		if len(parts) != 2 {
+			return assignedTags, fmt.Errorf("[%s] invalid tag, should be in format 'key:value' (only a single colon between the key and value, no spaces or special characters)", tag)
+		}
+		key := parts[0]
+		value := parts[1]
+		tagInput[key] = value
+	}
+	// assign tags listed in Terraform config but not found in service
+	assignedTags, err = client.AssignTags(string(resourceType.ResourceId()), tagInput)
```
rocktavious, line 247: Should we add an additional method maybe `AssignTagsPro` that takes in the more structured `[]Tag` 🤔   Then the current `AssignTags` could defer to it and basically just be a function that converts `map[string]string` to `[]Tag` and calls `AssignTagsPro`

rocktavious (commented):
Overall I think it gets a working process but in a clunky manner that would be improved with some additional code changes and maybe re-thinking the interface bit.   As it currently stands it doesn't feel very idomatic go and feels more like a script.

I do think its important to get this correct becuase opslevel-go has become more widely used directly by customers and doesn't just serve the CLI and Terraform anymore.  There are 3 customers i know of directly that ONLY use opslevel-go and we need to make sure we are giving them a good client experience too.

aliases.go
davidbloss: There is a `client.DeleteAlias()` and a `client.DeleteTag()` - but only DeleteAlias had its own wrapper functions like `DeleteServiceAlias` from 2021 and `DeleteInfraAlias` from 2023. I saw 2 paradigms to begin with and ran with the DeleteAlias funcs to match what already existed. We can walk these back and mark the wrapper Delete funcs as deprecated

PR #423, opened 2024-06-19
Db/rework reconcile funcs

davidbloss:
## Issues

Fix recently added reconcile funcs. 

## Changelog

- [X] List your changes here
- [ ] Make a `changie` entry

## Tophatting

`task test`

### DIFF
--- a/.changes/unreleased/Feature-20240624-103252.yaml
+++ b/.changes/unreleased/Feature-20240624-103252.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: add ReconcileAliases member function to System
+time: 2024-06-24T10:32:52.064724-05:00
--- a/.changes/unreleased/Feature-20240624-103252.yaml
+++ b/.changes/unreleased/Feature-20240624-103252.yaml
@@ -1,3 +1,0 @@
-kind: Feature
-body: add ReconcileAliases member function to System
-time: 2024-06-24T10:32:52.064724-05:00
--- a/.changes/unreleased/Feature-20240624-103433.yaml
+++ b/.changes/unreleased/Feature-20240624-103433.yaml
@@ -1,3 +1,3 @@
 kind: Feature
-body: add ReconcileAliases member function to Service
+body: add ReconcileAliases member function to Domain, InfrastructureResource, Scorecard, Service, System, and Team
 time: 2024-06-24T10:34:33.952247-05:00
--- a/.changes/unreleased/Feature-20240624-103433.yaml
+++ b/.changes/unreleased/Feature-20240624-103433.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: add ReconcileAliases member function to Service
+time: 2024-06-24T10:34:33.952247-05:00
--- a/aliases.go
+++ b/aliases.go
@@ -95,15 +95,17 @@ func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliases
 }
 
 // Given actual aliases and wanted aliases, returns aliasesToCreate and aliasesToDelete lists
-func extractAliases(existingAliases, aliasesWanted []string) ([]string, []string) {
+func ExtractAliases(existingAliases, aliasesWanted []string) ([]string, []string) {
 	var aliasesToCreate, aliasesToDelete []string
 
+	// collect aliasesToDelete - existing aliases that are no longer wanted
 	for _, alias := range existingAliases {
 		if !slices.Contains(aliasesWanted, alias) {
 			aliasesToDelete = append(aliasesToDelete, alias)
 		}
 	}
 
+	// collect aliasesToCreate - wanted aliases that do not yet exist
 	for _, aliasWanted := range aliasesWanted {
 		if !slices.Contains(existingAliases, aliasWanted) {
 			aliasesToCreate = append(aliasesToCreate, aliasWanted)
--- a/aliases.go
+++ b/aliases.go
@@ -7,7 +7,7 @@ import (
 )
 
 type AliasOwnerInterface interface {
-	ReconcileAliases(*Client, []string) ([]string, error)
+	ReconcileAliases(*Client, []string) error
 }
 
 func (client *Client) CreateAliases(ownerId ID, aliases []string) ([]string, error) {
@@ -89,8 +89,8 @@ func (client *Client) DeleteAlias(input AliasDeleteInput) error {
 
 // ReconcileAliases manages aliases API operations for AliasOwnerInterface implementations
 //
-// Aliases not in 'aliasesWanted' will be deleted, new tags from 'aliasesWanted' will be created. Reconciled aliases are returned.
-func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliasesWanted []string) ([]string, error) {
+// Aliases not in 'aliasesWanted' will be deleted, new tags from 'aliasesWanted' will be created
+func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliasesWanted []string) error {
 	return resourceType.ReconcileAliases(client, aliasesWanted)
 }
 
--- a/aliases.go
+++ b/aliases.go
@@ -7,7 +7,7 @@ import (
 )
 
 type AliasOwnerInterface interface {
-	AliasOwnerType() AliasOwnerTypeEnum
+	ReconcileAliases(*Client, []string) ([]string, error)
 }
 
 func (client *Client) CreateAliases(ownerId ID, aliases []string) ([]string, error) {
@@ -49,13 +49,7 @@ func (client *Client) CreateAlias(input AliasCreateInput) ([]string, error) {
 	return output, HandleErrors(err, m.Payload.Errors)
 }
 
-func (client *Client) DeleteDomainAlias(alias string) error {
-	return client.DeleteAlias(AliasDeleteInput{
-		Alias:     alias,
-		OwnerType: AliasOwnerTypeEnumDomain,
-	})
-}
-
+// Deprecated: use client.DeleteAlias instead
 func (client *Client) DeleteInfraAlias(alias string) error {
 	return client.DeleteAlias(AliasDeleteInput{
 		Alias:     alias,
@@ -63,13 +57,7 @@ func (client *Client) DeleteInfraAlias(alias string) error {
 	})
 }
 
-func (client *Client) DeleteScorecardAlias(alias string) error {
-	return client.DeleteAlias(AliasDeleteInput{
-		Alias:     alias,
-		OwnerType: AliasOwnerTypeEnumScorecard,
-	})
-}
-
+// Deprecated: use client.DeleteAlias instead
 func (client *Client) DeleteServiceAlias(alias string) error {
 	return client.DeleteAlias(AliasDeleteInput{
 		Alias:     alias,
@@ -77,13 +65,7 @@ func (client *Client) DeleteServiceAlias(alias string) error {
 	})
 }
 
-func (client *Client) DeleteSystemAlias(alias string) error {
-	return client.DeleteAlias(AliasDeleteInput{
-		Alias:     alias,
-		OwnerType: AliasOwnerTypeEnumSystem,
-	})
-}
-
+// Deprecated: use client.DeleteAlias instead
 func (client *Client) DeleteTeamAlias(alias string) error {
 	return client.DeleteAlias(AliasDeleteInput{
 		Alias:     alias,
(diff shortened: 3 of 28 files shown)


aliases.go
```diff
@@ -49,41 +49,23 @@ func (client *Client) CreateAlias(input AliasCreateInput) ([]string, error) {
 	return output, HandleErrors(err, m.Payload.Errors)
 }
 
-func (client *Client) DeleteDomainAlias(alias string) error {
```
davidbloss, line 52: This was added yesterday, removing today. We don't want to propagate this pattern.

aliases.go
```diff
@@ -49,41 +49,23 @@ func (client *Client) CreateAlias(input AliasCreateInput) ([]string, error) {
...
-
+// Deprecated: use client.DeleteAlias instead
 func (client *Client) DeleteInfraAlias(alias string) error {
 	return client.DeleteAlias(AliasDeleteInput{
 		Alias:     alias,
 		OwnerType: AliasOwnerTypeEnumInfrastructureResource,
 	})
 }
 
-func (client *Client) DeleteScorecardAlias(alias string) error {
```
davidbloss, line 66: This was added yesterday, removing today. We don't want to propagate this pattern.

aliases.go
```diff
@@ -49,41 +49,23 @@ func (client *Client) CreateAlias(input AliasCreateInput) ([]string, error) {
...
-
+// Deprecated: use client.DeleteAlias instead
 func (client *Client) DeleteServiceAlias(alias string) error {
 	return client.DeleteAlias(AliasDeleteInput{
 		Alias:     alias,
 		OwnerType: AliasOwnerTypeEnumService,
 	})
 }
 
-func (client *Client) DeleteSystemAlias(alias string) error {
```
davidbloss, line 80: This was added yesterday, removing today. We don't want to propagate this pattern.

tags.go
```diff
@@ -210,7 +223,7 @@ func (client *Client) DeleteTag(id ID) error {
 // ReconcileTags manages tags API operations for TaggableResourceInterface implementations
 //
 // Tags not in 'tagsWanted' will be deleted, new tags from 'tagsWanted' will be created. Reconciled tags are returned.
-func (client *Client) ReconcileTags(resourceType TaggableResourceInterface, tagsWanted []string) ([]Tag, error) {
+func (client *Client) ReconcileTags(resourceType TaggableResourceInterface, tagsWanted []Tag) ([]Tag, error) {
```
taimoorgit, line 226: `errors.Join()` should be used in this function too

tags_test.go
```diff
@@ -8,6 +8,105 @@ import (
...
+			expectedTagsToDelete: noTags,
+		},
+		"create none delete all": {
+			tagsWanted:           noTags,
+			existingTags:         fourTags,
+			expectedTagsToCreate: noTags,
+			expectedTagsToDelete: fourTags,
+		},
+		"create some delete some": {
+			tagsWanted:           fourTags[:3],
```
taimoorgit, line 41: IMO these tests are hard to read because of slice notation and also because I have to look up what `fourTags` is by scrolling up. I would prefer we just wrote the literal array in place of `noTags` and `fourTags`. Same for aliases_test.go.

(This is a non-blocking suggestion.)

taimoorgit (dismissed):

aliases_test.go
```diff
@@ -3,9 +3,59 @@ package opslevel_test
...
+			expectedAliasesToDelete: noAliases,
+		},
+		"create none delete all": {
+			aliasesWanted:           noAliases,
+			existingAliases:         fourAliases,
+			expectedAliasesToCreate: noAliases,
+			expectedAliasesToDelete: fourAliases,
+		},
+		"create some delete some": {
+			aliasesWanted:           fourAliases[:3],
```
davidbloss, line 35: Note to self: make this more readable

aliases.go
```diff
@@ -107,54 +89,27 @@ func (client *Client) DeleteAlias(input AliasDeleteInput) error {
 
 // ReconcileAliases manages aliases API operations for AliasOwnerInterface implementations
 //
-// Aliases not in 'aliasesWanted' will be deleted, new tags from 'aliasesWanted' will be created. Reconciled aliases are returned.
-func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliasesWanted []string) ([]string, error) {
-	var deleteAliasFunc func(string) error
-	var existingAliases []string
-	var resourceId ID
+// Aliases not in 'aliasesWanted' will be deleted, new tags from 'aliasesWanted' will be created
+func (client *Client) ReconcileAliases(resourceType AliasOwnerInterface, aliasesWanted []string) error {
```
rocktavious, line 93: Is there a reason to keep this around?

domain.go
```diff
@@ -24,8 +25,29 @@ type DomainConnection struct {
...
+		})
+		allErrors = errors.Join(allErrors, err)
+	}
+
+	if len(aliasesToCreate) > 0 {
+		// CreateAliases returns current list of aliases of Domain
+		d.ManagedAliases, err = client.CreateAliases(d.Id, aliasesToCreate)
+		allErrors = errors.Join(allErrors, err)
+	} else {
+		d.ManagedAliases = slices.DeleteFunc(d.ManagedAliases, func(alias string) bool {
```
rocktavious, line 45: I would think you would want to do this is regardless if len of aliasesToCreate > 0 ? 🤔

scorecards.go
```diff
@@ -28,16 +30,37 @@ type ScorecardConnection struct {
...
+	var allErrors, err error
+
+	aliasesToCreate, aliasesToDelete := ExtractAliases(scorecard.Aliases, aliasesWanted)
+	for _, alias := range aliasesToDelete {
+		err := client.DeleteAlias(AliasDeleteInput{
+			Alias:     alias,
+			OwnerType: AliasOwnerTypeEnumScorecard,
+		})
+		allErrors = errors.Join(allErrors, err)
+	}
```
rocktavious, line 49: This seems to be identical code 3 or more times - good candidate for a helper/utility function that takes in `aliasesToDelete` and `OwnerType` and returns `error`

tags.go
```diff
@@ -99,25 +102,35 @@ func ValidateTagKey(key string) error {
 	return nil
 }
 
-func (client *Client) AssignTags(identifier string, tags map[string]string) ([]Tag, error) {
+func (client *Client) AssignTagsWithTags(identifier string, tags []Tag) ([]Tag, error) {
```
rocktavious, line 105: Don't love the function name.  A customer with little context would read this and not really understand it.  This is why i recommended `AssignTagsPro` or `AssignTagsExt`.  You could even then add a function definition string like `// AssignTagsExt - Is AssignTags using a []Tag instead of map[string]string`

tags.go
```diff
@@ -99,25 +102,35 @@ func ValidateTagKey(key string) error {
...
+	var tagInput []TagInput
+	for _, tag := range tags {
+		tagInput = append(tagInput, TagInput{Key: tag.Key, Value: tag.Value})
+	}
```
rocktavious, line 109: This makes me think the function should actually take in `[]TagInput`

rocktavious (changes requested):
There is 1 thing that came up with a customer today.

It seems we've lost the ability to specify an alias that is one of the "auto gen" aliases for a given resource.  They get the following error

```
│ Unable to reconcile service aliases '[onezero-mt5 onezero-mt5_2]', go
│ error: OpsLevel API Errors:
│   - 'alias' Alias 'onezero-mt5_2' has already been taken
```

I think previously in 0.11.0 we would just exclude these aliases from reconciliation and ensure the rest work.

I'm not 100% sure if this code will fix this problem but we should make sure it does.

aliases.go
davidbloss: Not really. This can be dropped

domain.go
davidbloss: This definitely needs a comment explaining what's going on. 
`d.ManagedAliases` is updated via two paths. 
1. In the `if len(...) > 0` case, `client.CreateAliases()` returns the latest slice of aliases from the API which is stored in `d.ManagedAliases`. No local update needed.
2. In the `else` case,  `d.ManagedAliases` is updated manually since we're done with API calls here

tags.go
```diff
@@ -209,45 +222,55 @@ func (client *Client) DeleteTag(id ID) error {
...
-	var assignedTags []Tag
 
 	tagConnection, err = resourceType.GetTags(client, nil)
 	if err != nil {
-		return assignedTags, err
+		return err
 	}
 	if tagConnection == nil {
-		return assignedTags, fmt.Errorf("no tags found on %s with id '%s'", string(resourceType.ResourceType()), resourceType.ResourceId())
+		return fmt.Errorf("no tags found on %s with id '%s'", string(resourceType.ResourceType()), resourceType.ResourceId())
```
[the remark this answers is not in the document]
davidbloss: This is a good point and maybe a half baked error message on my side. But yes, I'll be updating this to better handle this really odd edge case

tags.go
```diff
@@ -209,45 +222,55 @@ func (client *Client) DeleteTag(id ID) error {
...
 	// delete tags found in resource but not listed in tagsWanted
-	for _, tag := range tagConnection.Nodes {
-		if !slices.Contains(tagsWanted, tag.Flatten()) {
-			if err := client.DeleteTag(tag.Id); err != nil {
-				return assignedTags, err
-			}
+	for _, tag := range tagsToDelete {
+		allErrors = errors.Join(allErrors, client.DeleteTag(tag.Id))
+	}
+	if allErrors != nil {
```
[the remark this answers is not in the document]
davidbloss: Ahh good catch!

domain_test.go
```diff
@@ -61,6 +61,84 @@ func TestDomainGetSystems(t *testing.T) {
...
+	aliasesWanted := []string{}
+	domain := ol.Domain{
+		DomainId: ol.DomainId{
+			Id: id1,
+		},
+		ManagedAliases: []string{"one", "two"},
+	}
+
+	// delete "one" alias
+	testRequestOne := autopilot.NewTestRequest(
```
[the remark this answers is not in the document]
taimoorgit: More detailed explanation of what I meant here after talking with @davidbloss 

1. `ExtractAliases()` is a really easily testable function because it returns the list of aliases that are to be deleted, and the list of aliases that are to be added.
2. `ReconcileAliases()` just does `ExtractAliases()` + `client.DeleteAlias()` and `client.AddAlias()` multiple times. This means that `ExtractAliases()` is kind of redundant and should at least be made private since it has no use outside of reconciliation.
3. `ReconcileAliases()` should be extremely explicit about what it did. It should not just return `errors` but also return a list of aliases that were added, aliases that were deleted. This way we know 100.0% that it is **making the correct decision to add and delete certain aliases** as part of reconciliation, and **we also know that the API calls it is making are valid API calls.**
4. Another way to solve this would be to add tooling to `BestTestClient` or the `autopilot` package that would let us setup a more explicit test expectation like:

```
client = newclient()
client.expects_calls = [{ full json for the first API call to be made }, { full json for the second API call to be made }]
domain.ReconcileAliases(client, wantedAliases)

# if client does not receive 2 calls, the test will fail
# if client does not receive call 1 and call 2 with exactly the expected json request defined in the test the test should fail.
```

^These types of test expectations can be set in Ruby test mocking libraries like `mocha`. We can do the same in Go but we need to write more code to support it.

---

I am lifting my 'request changes' for this because

1. I am certain the way the code is now that the API calls being made during reconciliation are correct. But the implementation is still subject to change.
2. There are MANY parts of the codebase where we don't have explicit expectations set like mentioned in (4).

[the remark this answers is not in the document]
taimoorgit: To add to (3), you can take this test and add on `testRequestThree` and `testRequestFour` and so on and it will still pass, which would be incorrect API requests for the reconcile function to make but the test would still pass because it is just checking that the result is correct.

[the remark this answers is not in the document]
taimoorgit: We can only confirm with `autopilot` that:

* That up to N API calls are made
* That if an API call is made, it is one that has been "registered" with autopilot.
* It does not account for less than N API calls being made or for the same API call being made several times. Or for an API call that was "registered" not being made. The test expectations outside of the API call logic have to deal with that.

tags.go
```diff
@@ -32,6 +31,10 @@ type Tag struct {
 	Value string `json:"value"`
 }
 
+func (t Tag) HasSameKeyValue(otherTag Tag) bool {
```
rocktavious, line 34: why not just `Equals` for the function name?

.changes/unreleased/Feature-20240624-103252.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Feature
+body: add ReconcileAliases member function to System
```
rocktavious, line 2: In the future you could just collapse all of these down into 1 change entry - especially if you are going to put them all into the same PR

tags.go
davidbloss: Because this does not account for the `Id`

export_test.go
```diff
@@ -0,0 +1,11 @@
+package opslevel
+
+// Workaround for testing unexported functions.
+//
+// Running `go help build` displays:
+// When compiling packages, build ignores files that end in '_test.go'.
```
rocktavious, line 6: 🤯  TIL

davidbloss: Same!

tags.go
davidbloss: When somebody provides "tags wanted" for creation, the Id will not be known for those tags

.changes/unreleased/Feature-20240624-103433.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Feature
```
rocktavious, line 1: I think we are missing a change entry for the tag reconciliation?

davidbloss: Good catch! 😁 - I'll add one

tags.go
```diff
@@ -209,45 +216,64 @@ func (client *Client) DeleteTag(id ID) error {
...
 	}
-	// assign tags listed in Terraform config but not found in service
-	assignedTags, err = client.AssignTags(string(resourceType.ResourceId()), tagInput)
-	if err != nil {
-		return assignedTags, err
+	return tagIdsToDelete
+}
+
+// return TagInputs for tags that are to be created
+func extractTagInputsToCreate(existingTags, tagsWanted []Tag) []TagInput {
```
rocktavious, line 265: I see now why we probably shouldn't have had the Assign function take in `TagInput` as it makes the extract functions not symmetrical.

rocktavious (approved):

PR #425, opened 2024-06-25
Db/update reconcile aliases funcs

davidbloss:
## Changelog

Update ReconcileAliases functions:
- Added GetResource() calls to ReconcileAliases(). Result is used to update `<resource>.Aliases` and `<resource>.ManagedAliases` if applicable
- Errors are batched together and returned at end of function 

- [X] List your changes here
- [ ] Make a `changie` entry

## Tophatting

`task test`

### DIFF
--- a/domain_test.go
+++ b/domain_test.go
@@ -66,7 +66,8 @@ func TestDomainReconcileAliasesDeleteAll(t *testing.T) {
 	aliasesWanted := []string{}
 	domain := ol.Domain{
 		DomainId: ol.DomainId{
-			Id: id1,
+			Id:      id1,
+			Aliases: []string{"one", "two"},
 		},
 		ManagedAliases: []string{"one", "two"},
 	}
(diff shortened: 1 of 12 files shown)


domain_test.go
```diff
@@ -83,14 +84,21 @@ func TestDomainReconcileAliasesDeleteAll(t *testing.T) {
 		`{"input":{ "alias": "two", "ownerType": "domain" }}`,
 		`{"data": { "aliasDelete": {"errors": [] }}}`,
 	)
-	requests := []autopilot.TestRequest{testRequestOne, testRequestTwo}
+	// get domain
+	testRequestThree := autopilot.NewTestRequest(
```
rocktavious, line 88: oh neat - bet that was unexpected failure but pretty cool.

rocktavious (approved):
Assuming we get the results we want with the Terraform provider the code looks 👍 to me

PR #429, opened 2024-07-03
update enums

davidbloss:
## Issues

[Fixes this](https://github.com/OpsLevel/terraform-provider-opslevel/issues/391)

## Changelog

Update enums

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task ci` works

### DIFF
--- a/enum.go
+++ b/enum.go
@@ -284,7 +284,6 @@ const (
 	CheckTypeGitBranchProtection CheckType = "git_branch_protection" // Verifies that all the repositories on the service have branch protection enabled.
 	CheckTypeServiceDependency   CheckType = "service_dependency"    // Verifies that the service has either a dependent or dependency.
 	CheckTypeRepoGrep            CheckType = "repo_grep"             // Run a comprehensive search across the service's repository using advanced search parameters.
-	CheckTypePackageVersion      CheckType = "package_version"       // Verifies certain aspects of a service using or not using software packages.
 )
 
 // All CheckType as []string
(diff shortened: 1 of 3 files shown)


enum.go
```diff
@@ -270,6 +284,7 @@ const (
 	CheckTypeGitBranchProtection CheckType = "git_branch_protection" // Verifies that all the repositories on the service have branch protection enabled.
 	CheckTypeServiceDependency   CheckType = "service_dependency"    // Verifies that the service has either a dependent or dependency.
 	CheckTypeRepoGrep            CheckType = "repo_grep"             // Run a comprehensive search across the service's repository using advanced search parameters.
+	CheckTypePackageVersion      CheckType = "package_version"       // Verifies certain aspects of a service using or not using software packages.
```
rocktavious, line 287: can you leave this one off for now - we don't support this check yet

enum.go
```diff
@@ -868,16 +972,20 @@ var AllToolCategory = []string{
 type UserRole string
 
 const (
-	UserRoleUser      UserRole = "user"       // A regular user on the account.
-	UserRoleAdmin     UserRole = "admin"      // An administrator on the account.
-	UserRoleBasicUser UserRole = "basic_user" // A basic user on the account with limited access.
+	UserRoleUser           UserRole = "user"            // A regular user on the account.
+	UserRoleAdmin          UserRole = "admin"           // An administrator on the account.
+	UserRoleBasicUser      UserRole = "basic_user"      // A basic user on the account with limited access.
```
rocktavious, line 977: `basic_user` is no more - why is it still here?

enum.go
```diff
@@ -462,6 +478,94 @@ var AllHasDocumentationTypeEnum = []string{
 	string(HasDocumentationTypeEnumAPI),
 }
 
+// PackageConstraintEnum represents possible values of a package version check constraint.
+type PackageConstraintEnum string
```
rocktavious, line 482: this one should be left out till we do the package check work in next sprint

enum.go
davidbloss: Good question, I'll remove it

taimoorgit (approved):
Should also add a `Removed` entry for `UserRoleBasicUser`/`basic_user`.

PR #433, opened 2024-07-19
update struct tags for fields in ServiceUpdateInputV2 struct

davidbloss:
## Issues

[Regression with update service command](https://github.com/OpsLevel/cli/issues/302)

## Changelog

- [x] List your changes here
- [ ] Make a `changie` entry, N/A internal 

## Tophatting

`task ci` lint + tests pass

### DIFF
--- a/.changes/unreleased/Refactor-20240719-135145.yaml
+++ b/.changes/unreleased/Refactor-20240719-135145.yaml
@@ -1,3 +1,3 @@
 kind: Refactor
-body: updated json and yaml struct tags on ServiceUpdateInput and ServiceUpdateInputV2 structs
+body: reverted change json and yaml struct tags on ServiceUpdateInput and ServiceUpdateInputV2 structs to what they were previously when ServiceUpdateInput was initially created and to ensure they match
 time: 2024-07-19T13:51:45.789647-05:00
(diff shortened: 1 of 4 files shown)


service_test.go
```diff
@@ -266,7 +266,8 @@ func TestCreateServiceWithParentSystem(t *testing.T) {
 
 func TestUpdateService(t *testing.T) {
 	addVars := `{"input":{"description": "The quick brown fox", "framework": "django", "id": "123456789", "lifecycleAlias": "pre-alpha", "name": "Hello World", "parent": {"alias": "some_system"}, "tierAlias": "tier_4"}}`
```
rocktavious, line 268: 🤔 should we revert v1 too since the codeGen messed it up?

rocktavious (changes requested):
This need a change entry because its changing a external contract back to what it was prior to the codegen

service_test.go
davidbloss: Yeah I think so. I think it would be confusing to have both with differing struct tags

.changes/unreleased/Refactor-20240719-135145.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Refactor
+body: updated json and yaml struct tags on ServiceUpdateInput and ServiceUpdateInputV2 structs
```
rocktavious, line 2:
```suggestion
body: reverted change json and yaml struct tags on ServiceUpdateInput and ServiceUpdateInputV2 structs to what they were previously when ServiceUpdateInput was initially created and to ensure they match
```

rocktavious (approved):

PR #435, opened 2024-07-23
improve filter predicate validation

davidbloss:
## Changelog

❗Upgrade module to use `go 1.22`
Add predicate validation to correct errors with predicates before sending API requests

- [x] List your changes here
- [x] Make a `changie` entry

## Tophatting

Tested with TF provider ✅

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -27,6 +27,7 @@ type FilterPredicate struct {
 	CaseSensitive *bool             `json:"caseSensitive,omitempty" yaml:"caseSensitive,omitempty" default:"false"`
 }
 
+// Validate the FilterPredicate based on known expectations before sending to API
 func (filterPredicate *FilterPredicate) Validate() error {
 	// validation common to Predicate and FilterPredicate types
 	basicPredicate := Predicate{Type: filterPredicate.Type, Value: filterPredicate.Value}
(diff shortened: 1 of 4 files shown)


filters.go
```diff
@@ -43,12 +62,95 @@ func (filterPredicate *FilterPredicate) Validate() error {
...
-	if filterPredicate.CaseSensitive != nil &&
-		*filterPredicate.CaseSensitive &&
+	if *filterPredicate.CaseSensitive &&
 		!slices.Contains(caseSensitiveTypes, filterPredicate.Type) {
 		return fmt.Errorf("FilterPredicate type '%s' cannot have CaseSensitive value set.", filterPredicate.Type)
 	}
+	return nil
+}
 
+func (filterPredicate *FilterPredicate) validateKeyHasExpectedType() error {
```
rocktavious, line 72: How do we intend to keep this up to date with the application?

rocktavious (approved):
While i'm not against this - it now becomes a burden to maintain the bespoke logic the monolith has over in a separate codebase.

Why is relying on the API errors not good enough?  Is that really worth the support burden of this logic when 6-12 months from now we have to change this because they've added new stuff? or they have removed one of these validations but our tools still reject it outright?

davidbloss:
> While i'm not against this - it now becomes a burden to maintain the bespoke logic the monolith has over in a separate codebase.
> 
> Why is relying on the API errors not good enough? Is that really worth the support burden of this logic when 6-12 months from now we have to change this because they've added new stuff? or they have removed one of these validations but our tools still reject it outright? (my gut says the is no value great enough for those edgecases but i leave that decision to you @davidbloss since you maintain the code more then me now)

It's the lesser of two evils really. The API errors themselves aren't always clear.
When trying to create a filter with a predicate that matches another filter, the error the API returns is
```bash
Unable to create filter, got error: OpsLevel API Errors:
      - '' Invalid filter ID
```
Similarly, providing a different invalid predicate returns
```bash
Unable to create filter, got error: OpsLevel API Errors:
      - 'value' Invalid global id 'HI'
```
The API error multiple customers have complained the most about is this one:
```bash
Unable to create filter, got error: OpsLevel API Errors:
      - 'type' invalid type for 'repository_ids' predicate'
```
☝️ there is no mention of the predicate that caused the error or indication of what predicate type should be used.

Long term I hope for a better solution on the API side but for now I think the added burden is worth it.

PR #447, opened 2024-08-07
Fix Infrastructure resource list graphql operation names

jasonopslevel:
## Changelog

- [ ]  updated the operation name of the graphql requests. Noticed this scanning datadog. These names should be unique to help us identify costly requests. 
- [ ] Make a `changie` entry <- Should I still add one? This one is pretty transparent and only matters to us I believe.


jasonopslevel:
> @jasonopslevel `IntegrationList` --> `InfrastructureResourceSchemaList` also needed in `integration_test.go`, `integration.go`, and `cache_test.go`

`integration.go` and  `integration_test.go` are correct (the graphql operation name _should_ be `IntegrationList` in that case, but I did update one instance `cache_test.go`.

davidbloss (approved):
LGTM! 🎉

PR #450, opened 2024-08-08
add GCP integration support

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/419

## Changelog

- [x] add `task gen-internal` that can generate the internal schema so we don't have to annoyingly go in and modify the client.go file every time
- [x] Add in input type, resource type
- [x] Add mutations
- [x] Make a `changie` entry

## Tophatting

See CLI PR - https://github.com/OpsLevel/cli/pull/312

### DIFF
--- a/integration.go
+++ b/integration.go
@@ -27,12 +27,18 @@ type AzureResourcesIntegrationFragment struct {
 	TenantId              string   `graphql:"tenantId"`
 }
 
+type GoogleCloudProject struct {
+	ID   string `graphql:"id"`
+	Name string `graphql:"name"`
+	URL  string `graphql:"url"`
+}
+
 type GoogleCloudIntegrationFragment struct {
-	Aliases               []string `graphql:"aliases"`
-	ClientEmail           string   `graphql:"clientEmail"`
-	OwnershipTagKeys      []string `graphql:"ownershipTagKeys"`
-	Projects              []string `graphql:"projects"`
-	TagsOverrideOwnership bool     `graphql:"tagsOverrideOwnership"`
+	Aliases               []string             `graphql:"aliases"`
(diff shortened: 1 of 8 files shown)


integration_test.go
```diff
@@ -78,13 +78,55 @@ func TestCreateAzureResourcesIntegration(t *testing.T) {
 	autopilot.Equals(t, "new azure resources", result.Name)
 	autopilot.Equals(t, "12345678-1234-1234-1234-123456789abc", result.TenantId)
 	autopilot.Equals(t, "12345678-1234-1234-1234-123456789def", result.SubscriptionId)
-	autopilot.Equals(t, []string{}, result.Aliases)
+	autopilot.Equals(t, []string{"new_azure_resources"}, result.AzureResourcesIntegrationFragment.Aliases)
```
taimoorgit, line 81: Updated this test to check for aliases because it's required by GoLang since the integration contains both "fragment" structs that contain an attribute named `Aliases`.

integration.go
```diff
@@ -27,6 +27,14 @@ type AzureResourcesIntegrationFragment struct {
 	TenantId              string   `graphql:"tenantId"`
 }
 
+type GoogleCloudIntegrationFragment struct {
+	Aliases               []string `graphql:"aliases"`
+	ClientEmail           string   `graphql:"clientEmail"`
+	OwnershipTagKeys      []string `graphql:"ownershipTagKeys"`
+	Projects              []string `graphql:"projects"`
```
davidbloss, line 34: 🤔 `Projects` in the API docs maps to a list of `GoogleCloudProject`

davidbloss (approved):
LGTM 🎉

PR #452, opened 2024-08-13
Remove `locked` from AlertSource, InfrastructureResource

taimoorgit:
## Issues

https://github.com/OpsLevel/opslevel-go/pull/434 -- correction for this PR, these two resources don't support a `locked` field and never have.

## Changelog

- [x] Remove field, ensure tests pass
- [x] Make a `changie` entry

### DIFF
--- a/.changes/unreleased/Bugfix-20240813-123252.yaml
+++ b/.changes/unreleased/Bugfix-20240813-123252.yaml
@@ -1,0 +1,4 @@
+kind: Bugfix
+body: Fix bug where AlertSource and InfraResource querires failed because of `locked`
+  field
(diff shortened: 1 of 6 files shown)


.changes/unreleased/Bugfix-20240813-123252.yaml
```diff
@@ -0,0 +1,4 @@
+kind: Bugfix
+body: Fix bug where AlertSource and InfraResource querires failed because of `locked`
```
rocktavious, line 2:
```suggestion
body: Fix bug where AlertSource and InfraResource queries failed because field name should be `ownerLocked` instead of `locked`
```

rocktavious (approved):

PR #453, opened 2024-08-15
Db/fitler out deactivated users

davidbloss:
## Issues

[As an End User, I can use filtering on the account users listing](https://github.com/OpsLevel/team-platform/issues/202)

## Changelog

Add filter query param to ListUsers() to filter out deactivated users

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

`task test`

### DIFF
--- a/user.go
+++ b/user.go
@@ -148,7 +148,7 @@ func (client *Client) ListUsers(variables *PayloadVariables) (*UserConnection, e
 	}
 	if variables == nil {
 		variables = client.InitialPageVariablesPointer()
-		(*variables)["filter"] = []UsersFilterInput{}
+		(*variables)["filter"] = &[]UsersFilterInput{}
 	}
 
 	if err := client.Query(&q, *variables, WithName("UserList")); err != nil {
--- a/user.go
+++ b/user.go
@@ -148,7 +148,7 @@ func (client *Client) ListUsers(variables *PayloadVariables) (*UserConnection, e
 	}
 	if variables == nil {
 		variables = client.InitialPageVariablesPointer()
-		(*variables)["filter"] = &[]UsersFilterInput{}
+		(*variables)["filter"] = new([]UsersFilterInput)
 	}
 
 	if err := client.Query(&q, *variables, WithName("UserList")); err != nil {
--- a/user.go
+++ b/user.go
@@ -143,11 +143,12 @@ func (client *Client) GetUser(value string) (*User, error) {
 func (client *Client) ListUsers(variables *PayloadVariables) (*UserConnection, error) {
 	var q struct {
 		Account struct {
-			Users UserConnection `graphql:"users(after: $after, first: $first)"`
+			Users UserConnection `graphql:"users(after: $after, first: $first, filter: $filter)"`
 		}
 	}
 	if variables == nil {
 		variables = client.InitialPageVariablesPointer()
+		(*variables)["filter"] = []UsersFilterInput{}
 	}
 
 	if err := client.Query(&q, *variables, WithName("UserList")); err != nil {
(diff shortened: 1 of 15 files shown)


user_test.go
```diff
@@ -103,6 +103,40 @@ func TestListUser(t *testing.T) {
...
+
+	client := BestTestClient(t, "user/list_omit_deactivated", requests...)
+	// Act
+	payloadVars := client.InitialPageVariablesPointer()
+	(*payloadVars)["filter"] = &[]ol.UsersFilterInput{
+		{
+			Key:  ol.UsersFilterEnumDeactivatedAt,
+			Type: ol.RefOf(ol.BasicTypeEnumEquals),
+		},
+	}
```
rocktavious, line 128: This is SUPER ugly and unfriendly to the end user.  Any way we can make this customer facing interface better?  Maybe functional options passed into `ListUsers`?

IE
```
response, err := client.ListUsers(nil, ol.WithoutDeactivedUsers())
```

user.go
```diff
@@ -143,11 +143,12 @@ func (client *Client) GetUser(value string) (*User, error) {
 func (client *Client) ListUsers(variables *PayloadVariables) (*UserConnection, error) {
 	var q struct {
 		Account struct {
-			Users UserConnection `graphql:"users(after: $after, first: $first)"`
+			Users UserConnection `graphql:"users(after: $after, first: $first, filter: $filter)"`
 		}
 	}
 	if variables == nil {
 		variables = client.InitialPageVariablesPointer()
+		(*variables)["filter"] = &[]UsersFilterInput{}
```
rocktavious, line 151: why a pointer to an empty slice?  wouldn't `[]UserFilterInput{}` work?  Also i think my golang always complains about slice initialization like this because the more idomatic go is `make([]UserFilerInput)` but maybe i'm miss remembering that.  I know it complains 1 way or the other.

davidbloss: A pointer to `[]UsersFilterInput` resolves to the nullable `[UsersFilterInput!]` (matching the API) whereas a concrete `[]UsersFilterInput` resolves to the non-nullable `[UsersFilterInput!]!`

I'll update this to `new([]UserFilerInput)` as this resolves to `filter: null` in the query

user_test.go
```diff
@@ -103,6 +103,34 @@ func TestListUser(t *testing.T) {
...
+	testRequestTwo := autopilot.NewTestRequest(
+		`query UserList($after:String!$filter:[UsersFilterInput!]$first:Int!){account{users(after: $after, first: $first, filter: $filter){nodes{id,email,htmlUrl,name,role},{{ template "pagination_request" }},totalCount}}}`,
+		`{ "after": "OA", "filter": [ {"key": "deactivated_at", "type": "equals"} ], "first": 100 }`,
+		`{ "data": { "account": { "users": { "nodes": [], {{ template "pagination_second_pageInfo_response" }}, "totalCount": 0 }}}}`,
+	)
+	requests := []autopilot.TestRequest{testRequestOne, testRequestTwo}
+
+	client := BestTestClient(t, "user/list_omit_deactivated", requests...)
+	// Act
+	payloadVars := client.InitialPageVariablesPointer().WithoutDeactivedUsers()
```
rocktavious, line 122: While this is not [functional options](https://golang.cafe/blog/golang-functional-options-pattern.html) - i don't hate it and it is better then before.

user.go
rocktavious: TIL

user_test.go
davidbloss: I started testing out the functional options approach but think the builder pattern is more appropriate in this case.
We already have `client.InitialPageVariables()` and `client.InitialPageVariablesPointer()`. Adding some form of `New()` without a client, or `New(client, ....)` seems like bloat

rocktavious (approved):

user_test.go
rocktavious: 👍

PR #455, opened 2024-08-21
Fetch external IDs for team contacts

romcheg:
## Changelog

-  Fetch external IDs for team contacts by default


rocktavious:
This doesn't actually change anything as you can see from this test we already generate the graphql query body for contacts as `externalId`. https://github.com/OpsLevel/opslevel-go/blob/main/team_test.go#L144

PR #456, opened 2024-08-21
tags with matching keys no longer lost on update

davidbloss:
## Issues

[Bug when adding new tags to an opslevel_service that already has existing tags](https://github.com/OpsLevel/terraform-provider-opslevel/issues/437)

## Changelog

Stop filtering existingTags out of tagsWanted during ReconcileTags(). 
When multiple tags with same key exist, they should all be passed to `tagAssign()`. As is, updating a service with tag `foo:bar` with both `foo:bar` and `foo:bar123` will not work. `foo:bar` already exists, is dropped, then tagAssign only includes `foo:bar123` which replaces `foo:bar`. These lost tags only occurs when the keys (e.g. `foo`) match

Remove `extractTagInputsToCreate()` as it's no longer needed

- [X] List your changes here
- [X] Make a `changie` entry

### DIFF
--- a/tags_test.go
+++ b/tags_test.go
@@ -61,49 +61,6 @@ func TestExtractTagIdsToDelete(t *testing.T) {
 	}
 }
 
-type extractTagInputsTestCase struct {
-	existingTags              []ol.Tag
-	expectedTagInputsToCreate []ol.TagInput
-	tagsWanted                []ol.Tag
-}
-
-func TestExtractTagInputsToCreate(t *testing.T) {
-	var noTagInputs []ol.TagInput
-	allTagInputs := []ol.TagInput{
-		{Key: tagOne.Key, Value: tagOne.Value},
-		{Key: tagTwo.Key, Value: tagTwo.Value},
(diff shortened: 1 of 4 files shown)


tags_test.go
```diff
@@ -61,49 +61,6 @@ func TestExtractTagIdsToDelete(t *testing.T) {
...
-
-			// Assert
-			autopilot.Equals(t, tagInputsToCreate, tc.expectedTagInputsToCreate)
-		})
-	}
-}
-
 type tagHasSameKeyValueTestCase struct {
 	tagOne          ol.Tag
 	tagTwo          ol.Tag
```
rocktavious, line 66: I think we should try to beef up these test cases a bit to account for more then what they currently do.

davidbloss: True and I'm all for beefing up tests. I think this might have been an odd one anyway since the behavior on the other side of the API is slightly different based on the tag keys. Knowing that now we can write tests for this though

davidbloss:
Redoing this with `tagCreate()` and `tagDelete()`

PR #460, opened 2024-08-26
Fix Reconcile Tags

rocktavious:
Resolves #

### Problem

Previously ReconcileTags was not properly reconciling the tags of a taggable resource, some tags weren't getting deleted and some tags were sticking around incorrectly.

### Solution

After deep discovery - we have determined that we need to use TagCreate and TagDelete directly and perform the reconciliation ourselves.  Additionally i've ensured via table tests `TestReconcileTags` that we cover all the different permutations of current tags and desired tags to ensure we perform the correct API calls.  Then i've added a test `TestReconcileTagsAPI` to ensure we actually make those API calls.

### Checklist

  - If so what is the ticket or PR # TBD

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -5,7 +5,6 @@ package opslevel
 // Running `go help build` displays:
 // When compiling packages, build ignores files that end in '_test.go'.
 var (
-	ExtractAliases           = extractAliases
-	ExtractTagIdsToDelete    = extractTagIdsToDelete
-	ExtractTagInputsToCreate = extractTagInputsToCreate
+	TestExtractAliases = extractAliases
+	TestReconcileTags  = reconcileTags
 )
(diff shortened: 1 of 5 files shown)


export_test.go
```diff
@@ -5,7 +5,6 @@ package opslevel
 // Running `go help build` displays:
 // When compiling packages, build ignores files that end in '_test.go'.
 var (
-	ExtractAliases           = extractAliases
-	ExtractTagIdsToDelete    = extractTagIdsToDelete
-	ExtractTagInputsToCreate = extractTagInputsToCreate
+	TestExtractAliases = extractAliases
+	TestReconcileTags  = reconcileTags
```
davidbloss, line 9: 👏 Great choice of renaming!

PR #464, opened 2024-09-05
add InviteUser with forceSendInvite bool and test

davidbloss:
Resolves [#463](https://github.com/OpsLevel/team-platform/issues/463)

### Checklist


user.go
```diff
@@ -127,6 +127,22 @@ func (client *Client) InviteUser(email string, input UserInput) (*User, error) {
 	return &m.Payload.User, HandleErrors(err, m.Payload.Errors)
 }
 
+func (client *Client) InviteUserWithForceSendInvite(email string, input UserInput, forceSendInvite *bool) (*User, error) {
```
rocktavious, line 130: Lets do a breaking change here instead.

rocktavious (changes requested):
I think this is one of those times where we need to do a breaking change.

user.go
davidbloss: 🫡 I was debating on what to do here

user.go
```diff
@@ -112,16 +112,17 @@ func (user *User) Teams(client *Client, variables *PayloadVariables) (*TeamIdCon
 	return &q.Account.User.Teams, nil
 }
 
-func (client *Client) InviteUser(email string, input UserInput) (*User, error) {
+func (client *Client) InviteUser(email string, input UserInput, forceSendInvite *bool) (*User, error) {
```
rocktavious, line 115:
```suggestion
func (client *Client) InviteUser(email string, input UserInput, sendInvite bool) (*User, error) {
```

Does it need to be a bool pointer?  Since we are now requiring the end user to specify during invite time we can just pass `true` or `false` - that would be nicer to the end user instead of having to `ol.RefOf(true)`

davidbloss: updated to be a bool

rocktavious (approved):
LGTM - This will be good ergonomics for end users now that it actually works sanely

.changes/unreleased/Refactor-20240905-144928.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Refactor
+body: 'BREAKING CHANGE: new forceSendInvite arg added to UserInvite()'
```
rocktavious, line 2:
```suggestion
body: 'BREAKING CHANGE: new argument sendInvite added to UserInvite() method to force send a user invite e-mail'
```

PR #470, opened 2024-10-04
add CodeIssue check type

davidbloss:
Resolves [#504](https://github.com/OpsLevel/team-platform/issues/504)

Added CodeIssue check and update check tests to be more robust

### Checklist


check_code_issue.go
```diff
@@ -0,0 +1,28 @@
+package opslevel
+
+type CodeIssueCheckFragment struct {
+	Campaign string `graphql:"campaign"` // The campaign the check belongs to.
```
rocktavious, line 4: This does not feel correct - why was this added?  (This feels like its a field on the base check type)

input.go
```diff
@@ -101,6 +101,43 @@ type CheckAlertSourceUsageUpdateInput struct {
...
+	LevelId        ID                            `json:"levelId" yaml:"levelId" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                       // The id of the level the check belongs to. (Required.)
+	OwnerId        *ID                           `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`   // The id of the team that owns the check. (Optional.)
+	FilterId       *ID                           `json:"filterId,omitempty" yaml:"filterId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // The id of the filter of the check. (Optional.)
+	Notes          *string                       `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_notes"`                         // Additional information about the check. (Optional.)
+	Constraint     CheckCodeIssueConstraintEnum  `json:"constraint" yaml:"constraint" example:"NEW_ENUM_SET_DEFAULT"`                            // The type of constraint used in evaluation the code issues check. (Required.)
```
rocktavious, line 114: The `example` tag is wrong?

input.go
```diff
@@ -595,6 +632,12 @@ type ChecksCopyToCampaignInput struct {
 	CheckIds   []string `json:"checkIds" yaml:"checkIds" example:"['Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk', 'Z2lkOi8vc2VydmljZS85ODc2NTQzMjE']"` // The IDs of the existing rubric checks to be copied. (Required.)
 }
 
+// CodeIssueResolutionTimeInput represents the allowed threshold for how long an issue has been detected before the check starts failing.
+type CodeIssueResolutionTimeInput struct {
+	Unit  CodeIssueResolutionTimeUnitEnum `json:"unit" yaml:"unit" example:"NEW_ENUM_SET_DEFAULT"` // . (Required.)
+	Value int                             `json:"value" yaml:"value" example:"3"`                  // . (Required.)
```
rocktavious, line 638: feels like we are missing docstrings on this - is that an API problem?

check_code_issue.go
rocktavious: Also where is `Constraint` and `IssueName` and `ResolutionTime` and `MaxAllowed` etc etc.

These are surely in the response payload.

check_code_issue.go
```diff
@@ -0,0 +1,28 @@
...
+	Url      string `graphql:"url"`      // The url to the check.
```
rocktavious, line 5: Same with this field - this is likely on the base check type not specific to the CodeIssuesCheck type.

rocktavious (changes requested):
Lots of issues with this one.  Let me know if you want to sit down and chat through it.

check_code_issue.go
davidbloss: Oddly enough they are not. I'll ask around and see if these were forgotten in the API

check_code_issue.go
davidbloss: You're right on. `Campaign` and `Url` are both part of the base check type

input.go
davidbloss: Yes and good call!

input.go
```diff
@@ -595,6 +632,12 @@ type ChecksCopyToCampaignInput struct {
 	CheckIds   []string `json:"checkIds" yaml:"checkIds" example:"['Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk', 'Z2lkOi8vc2VydmljZS85ODc2NTQzMjE']"` // The IDs of the existing rubric checks to be copied. (Required.)
 }
 
+// CodeIssueResolutionTimeInput represents the allowed threshold for how long an issue has been detected before the check starts failing.
+type CodeIssueResolutionTimeInput struct {
+	Unit  CodeIssueResolutionTimeUnitEnum `json:"unit" yaml:"unit" example:"NEW_ENUM_SET_DEFAULT"` // . (Required.)
```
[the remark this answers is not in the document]
davidbloss: Yes wrong here too

check_test.go
```diff
@@ -258,6 +278,73 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
...
+			expectedCheck: CheckWithExtras(map[string]any{
+				"constraint":     "exact",
+				"issueName":      "test-issue",
+				"issueType":      []string{"bug", "error"},
+				"maxAllowed":     3,
+				"resolutionTime": map[string]any{"unit": "day", "value": 1},
+				"severity":       []string{"sev1", "sev2"},
+			}),
+		},
+		"UpdateCodeIssue": {
```
rocktavious, line 318: I think we'll want to throw a few more "update" test cases in here to ensure it generates the right input for a few of these optional fields.  `maxAllowed`, `resolutionTime` etc etc

rocktavious (approved):
If the extra tests are too tricky to write correctly because of API validation - then this is 👍

check_code_issue.go
```diff
@@ -2,10 +2,10 @@ package opslevel
 
 type CodeIssueCheckFragment struct {
 	Constraint     CheckCodeIssueConstraintEnum `graphql:"constraint"`     // The type of constraint used in evaluation the code issues check.
-	IssueName      string                       `graphql:"issueName"`      // The issue name used for code issue lookup.
+	IssueName      *string                      `graphql:"issueName"`      // The issue name used for code issue lookup.
 	IssueType      []string                     `graphql:"issueType"`      // The type of code issue to consider.
-	MaxAllowed     int                          `graphql:"maxAllowed"`     // The threshold count of code issues beyond which the check starts failing.
-	ResolutionTime CodeIssueResolutionTime      `graphql:"resolutionTime"` // The resolution time recommended by the reporting source of the code issue.
+	MaxAllowed     *int                         `graphql:"maxAllowed"`     // The threshold count of code issues beyond which the check starts failing.
+	ResolutionTime *CodeIssueResolutionTime     `graphql:"resolutionTime"` // The resolution time recommended by the reporting source of the code issue.
```
rocktavious, line 8: What caused us to change these to pointers?

davidbloss: Classic whoopsie daisy

rocktavious (approved):
Heafty but good work!!

PR #471, opened 2024-10-07
add regionOverride field to AWS integration, for overriding multiple regions

davidbloss:
Resolves [#505](https://github.com/OpsLevel/team-platform/issues/505)

### Checklist

### DIFF
--- a/integration.go
+++ b/integration.go
@@ -17,6 +17,7 @@ type AWSIntegrationFragment struct {
 	ExternalID           string   `graphql:"externalId"`
 	OwnershipTagOverride bool     `graphql:"awsTagsOverrideOwnership"`
 	OwnershipTagKeys     []string `graphql:"ownershipTagKeys"`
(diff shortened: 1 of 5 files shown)


integration.go
```diff
@@ -53,11 +54,12 @@ type IntegrationConnection struct {
 }
 
 type AWSIntegrationInput struct {
```
rocktavious, line 56: It seems this struct is duplicative of whats in `input.go` ?  why?

rocktavious (approved):
LGTM - just not sure why we have duplicative "input" structs

integration.go
davidbloss: Yeah I noticed that... not sure what happened here but didn't want to yank out the duplicate before figuring out which is the "good" one

rocktavious: Lets make sure we follow this up

PR #498, opened 2024-12-26
generate input objects, start using Nullable

davidbloss:
Resolves [#573](https://github.com/OpsLevel/team-platform/issues/573)

### 🚨 This PR contains BREAKING CHANGES for the upcoming `v2025` opslevel-go release 🚨 

Run `task gen` to see new code generated in `input.go`

### Noteworthy Changes

- ❗The `*Nullable[<type>]` type now wraps optional fields of most structs in `input.go`. This sweeping change enables setting input object fields to JSON `null`, resolves ambiguity during API update calls, and completes a planned major refactor before the next major release of opslevel-go.
- In `scalar.go`, a new `UnmarshalJSON()` for `Nullable` types was added to correctly unmarshal JSON values to this custom type.
- `task gen` now uses the template in `templates/inputObjects.tpl` to generate everything in `input.go`
- In `gen.go`: `enumExamples`, `listExamples`, and `scalarExamples` are used to set values in `example:` struct tags
- All `RefOf()` calls have been replaced with `NewNullableFrom()`
- Tests have been updated that verify these changes.

### Checklist


actions_test.go
```diff
@@ -83,7 +83,7 @@ func TestUpdateWebhookAction2(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
 		`mutation WebhookActionUpdate($input:CustomActionsWebhookActionUpdateInput!){customActionsWebhookActionUpdate(input: $input){webhookAction{{ template "custom_actions_request" }},errors{message,path}}}`,
-		`{"input":{"id": "123456789","description":"","headers":"{\"Accept\":\"application/json\"}"}}`,
+		`{"input":{"id": "123456789","description":null,"headers":"{\"Accept\":\"application/json\"}"}}`,
```
rocktavious, line 86: This seems like a material change - i'm not sure we want this.

gen.go
```diff
@@ -110,6 +109,97 @@ var stringTypeSuffixes = []string{
 	"yaml",
 }
 
+var enumExamples = map[string]string{
+	"AlertSourceStatusTypeEnum":                       opslevel.AllAlertSourceStatusTypeEnum[0],
+	"ApiDocumentSourceEnum":                           opslevel.AllApiDocumentSourceEnum[0],
+	"AlertSourceTypeEnum":                             opslevel.AllAlertSourceTypeEnum[0],
```
rocktavious, line 115: This feels like a chicken and egg problem since these are generated data - how does this work when enum.go doesn't exist?

rocktavious: Reading further down - i think you can just replace this with data from the schema if you build it up during schema parsing first.

gen.go
```diff
@@ -891,6 +1001,161 @@ func renameMutation(s string) string {
...
+	case "Float":
+		convertedType += "float64"
+	case "Int":
+		convertedType += "int"
+	case "ISO8601DateTime":
+		convertedType += "iso8601.Time"
+	case "String":
+		convertedType += "string"
+	default:
+		convertedType += graphqlType
```
rocktavious, line 1097: should we log a warning when this happens so we can be alerted to it?

gen.go
```diff
@@ -891,6 +1001,161 @@ func renameMutation(s string) string {
...
+			goType = "*JsonString"
+		case "PropertyDefinition":
+			goType = "PropertyDefinitionId"
+		}
+	} else if fieldType.Name == "ServiceRepository" && goType == "Repository" {
+		goType = "RepositoryId"
+	}
+
+	return goType
+}
```
rocktavious, line 1153: This logic is pretty hard to follow.  Something to mark for refactoring or thinking if there is a better way.  I don't have a good handle on all the context needed to understand this yet but feels like a missing design pattern we could leverage that would significantly clean this up.  Maybe the Open/Closed principle or the Liskov Substitution principle.

input.go
```diff
@@ -24,41 +24,108 @@ type AlertSourceServiceDeleteInput struct {
...
+	SubscriptionId        *Nullable[string]   `json:"subscriptionId,omitempty" yaml:"subscriptionId,omitempty" example:"example_value"`                // The subscription OpsLevel uses to access the Azure account. (Optional.)
+	TagsOverrideOwnership *Nullable[bool]     `json:"tagsOverrideOwnership,omitempty" yaml:"tagsOverrideOwnership,omitempty" example:"false"`          // Allow tags imported from Azure to override ownership set in OpsLevel directly. (Optional.)
+	TenantId              *Nullable[string]   `json:"tenantId,omitempty" yaml:"tenantId,omitempty" example:"example_value"`                            // The tenant OpsLevel uses to access the Azure account. (Optional.)
+}
+
+// CampaignCreateInput specifies the input fields used to create a campaign.
+type CampaignCreateInput struct {
+	CheckIdsToCopy *Nullable[[]ID]                  `json:"checkIdsToCopy,omitempty" yaml:"checkIdsToCopy,omitempty" example:"[]"`                  // The IDs of the existing rubric checks to be copied. (Optional.)
```
rocktavious, line 60: I think this might be an edgecase of not really truely being "optional". Also i'm not sure nullable of a slice makes sense.

input.go
```diff
@@ -24,41 +24,108 @@ type AlertSourceServiceDeleteInput struct {
...
+	OwnershipTagKeys      *Nullable[[]string] `json:"ownershipTagKeys,omitempty" yaml:"ownershipTagKeys,omitempty" example:"['tag_key1', 'tag_key2']"` // An array of tag keys used to associate ownership from an integration. Max 5. (Optional.)
+	SubscriptionId        *Nullable[string]   `json:"subscriptionId,omitempty" yaml:"subscriptionId,omitempty" example:"example_value"`                // The subscription OpsLevel uses to access the Azure account. (Optional.)
+	TagsOverrideOwnership *Nullable[bool]     `json:"tagsOverrideOwnership,omitempty" yaml:"tagsOverrideOwnership,omitempty" example:"false"`          // Allow tags imported from Azure to override ownership set in OpsLevel directly. (Optional.)
+	TenantId              *Nullable[string]   `json:"tenantId,omitempty" yaml:"tenantId,omitempty" example:"example_value"`                            // The tenant OpsLevel uses to access the Azure account. (Optional.)
+}
+
+// CampaignCreateInput specifies the input fields used to create a campaign.
+type CampaignCreateInput struct {
```
rocktavious, line 59: We should probably put in an exclusion for everything "campaigns" related.  The API has never been vetted throughly for external use.  Its entirely possible that these structures don't work correctly.

input.go
```diff
@@ -6,15 +6,15 @@ import "github.com/relvacode/iso8601"
 
 // AlertSourceExternalIdentifier specifies the input needed to find an alert source with external information.
 type AlertSourceExternalIdentifier struct {
-	Type       AlertSourceTypeEnum `json:"type" yaml:"type" example:"pagerduty"`                                   // The type of the alert. (Required.)
-	ExternalId string              `json:"externalId" yaml:"externalId" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // The external id of the alert. (Required.)
+	ExternalId string              `json:"externalId" yaml:"externalId" example:"example_value"` // The external id of the alert. (Required.)
```
rocktavious, line 9: why is the example value in the struct tag changing?

davidbloss: This is an edge case that needs special handling. The `Z2lk...` value is populated for fields of type `Id`. In the API schema this is string but I believe we expect it to have a value like an Id under the hood.

input.go
davidbloss: This is definitely an odd case, but from a purely JSON perspective it makes sense. Unless some JSON schema forbids it, `null` fits pretty much anywhere. The mystery then is how it's handled on the app side

gen.go
davidbloss: I'll work on this

gen.go
davidbloss: I don't think a warning is needed here. The `default` case catches non-scalar API Schema defined types like `IdentifierInput`, `UserIdentifierInput`, `ToolCategory`, `AlertSourceExternalIdentifier`, etc

gen.go
davidbloss: I've got a simple fix for this 👍

input.go
davidbloss: Campaign related objects no longer generated ✅

gen.go
davidbloss: This function is not used in this PR. Resolving this thread.

actions.go
```diff
@@ -165,10 +165,12 @@ func (client *Client) CreateTriggerDefinition(input CustomActionsTriggerDefiniti
 		} `graphql:"customActionsTriggerDefinitionCreate(input: $input)"`
 	}
 	if input.AccessControl == nil {
-		input.AccessControl = RefOf(CustomActionsTriggerDefinitionAccessControlEnumEveryone)
+		everyoneEnum := CustomActionsTriggerDefinitionAccessControlEnumEveryone
+		input.AccessControl = &everyoneEnum
```
rocktavious, line 169: Did you test if changing the enum to `var` instead of `const` makes is so we don't have to do this?

It should be a transparent change to the end user and make it so you can just do

```suggestion
		input.AccessControl = &CustomActionsTriggerDefinitionAccessControlEnumEveryone
```

davidbloss: That does the trick. Should I open a separate PR to make this change to enum.go?

check_test.go
```diff
@@ -346,31 +349,31 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
 			fixture: BuildUpdateRequest("CodeIssue", map[string]any{
 				"constraint":     ol.CheckCodeIssueConstraintEnumContains,
 				"issueName":      "code-issue-updated",
-				"issueType":      nil, // API NOTE - not allowed when constraint is "contains"
+				"issueType":      &[]string{}, // API NOTE - not allowed when constraint is "contains"
```
rocktavious, line 352: shouldn't this be `(*[]string)(nil)` to be the same as before?  I guess i would expect `&[]string{}` to be an initialized by empty slice and thus generating `[]` ?

actions.go
davidbloss: Done ✅

check_test.go
```diff
@@ -404,7 +407,7 @@ func getCheckTestCases() map[string]TmpCheckTestCase {
 			fixture: BuildUpdateRequest("HasRecentDeploy", map[string]any{"days": 5}),
 			body: func(c *ol.Client) (*ol.Check, error) {
 				input := ol.NewCheckUpdateInputTypeOf[ol.CheckHasRecentDeployUpdateInput](checkUpdateInput)
-				input.Days = ol.RefOf(5)
+				input.Days = ol.NewNullableFrom(5)
```
rocktavious, line 408: heres an `int` where i'm not sure `null` makes sense - can we test the API in Graphiql?

actions_test.go
```diff
@@ -69,9 +71,10 @@ func TestUpdateWebhookAction(t *testing.T) {
 	client := BestTestClient(t, "custom_actions/update_action", testRequest)
 
 	// Act
+	putHttpMethod := ol.CustomActionsHttpMethodEnumPut
```
rocktavious, line 74: missed

common.go
```diff
@@ -22,11 +22,12 @@ type PayloadVariables map[string]interface{}
 
 // WithoutDeactivedUsers filters out deactivated users on ListUsers query
 func (pv *PayloadVariables) WithoutDeactivedUsers() *PayloadVariables {
+	enumEquals := BasicTypeEnumEquals
```
rocktavious, line 25: missed

actions_test.go
```diff
@@ -186,12 +190,12 @@ func TestCreateTriggerDefinitionWithNullExtendedTeams(t *testing.T) {
 	trigger, err := client.CreateTriggerDefinition(ol.CustomActionsTriggerDefinitionCreateInput{
 		Name:                   "Deploy Rollback",
 		Description:            ol.RefOf("Disables the Deploy Freeze"),
-		AccessControl:          ol.RefOf(ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone),
+		AccessControl:          &ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone,
 		ManualInputsDefinition: ol.RefOf(""),
 		ResponseTemplate:       ol.RefOf(""),
-		ActionId:               newID,
+		ActionId:               ol.RefOf(*newID),
```
rocktavious, line 196: this is uncessary?

actions_test.go
```diff
@@ -130,13 +134,13 @@ func TestCreateTriggerDefinition(t *testing.T) {
 	// Act
 	trigger, err := client.CreateTriggerDefinition(ol.CustomActionsTriggerDefinitionCreateInput{
 		Name:                   "Deploy Rollback",
-		AccessControl:          ol.RefOf(ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone),
+		AccessControl:          &ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone,
 		Description:            ol.RefOf("Disables the Deploy Freeze"),
 		ResponseTemplate:       ol.RefOf(""),
 		ManualInputsDefinition: ol.RefOf(""),
-		ActionId:               newID,
+		ActionId:               ol.RefOf(*newID),
```
rocktavious, line 141: unecessary?

actions_test.go
```diff
@@ -155,14 +159,14 @@ func TestCreateTriggerDefinitionWithGlobalEntityType(t *testing.T) {
 	// Act
 	trigger, err := client.CreateTriggerDefinition(ol.CustomActionsTriggerDefinitionCreateInput{
 		Name:                   "Deploy Rollback",
-		AccessControl:          ol.RefOf(ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone),
+		AccessControl:          &ol.CustomActionsTriggerDefinitionAccessControlEnumEveryone,
 		Description:            ol.RefOf("Disables the Deploy Freeze"),
-		ActionId:               newID,
+		ActionId:               ol.RefOf(*newID),
```
rocktavious, line 164: unecessary?

check_test.go
```diff
@@ -32,8 +32,9 @@ var (
 		Value: ol.RefOf("Requests"),
 	}
 
+	equalsEnum           = ol.PredicateTypeEnumEquals
```
rocktavious, line 35: missed?

input.go
```diff
@@ -68,586 +68,609 @@ type CategoryDeleteInput struct {
...
+	MaxAllowed     *Nullable[int]                `json:"maxAllowed,omitempty" yaml:"maxAllowed,omitempty" example:"3"`                               // The threshold count of code issues beyond which the check starts failing. (Optional.)
+	Name           *Nullable[string]             `json:"name,omitempty" yaml:"name,omitempty" example:"example_value"`                               // The display name of the check. (Optional.)
+	Notes          *Nullable[string]             `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                             // Additional information about the check. (Optional.)
+	OwnerId        *Nullable[ID]                 `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`       // The id of the owner of the check. (Optional.)
+	ResolutionTime *CodeIssueResolutionTimeInput `json:"resolutionTime,omitempty" yaml:"resolutionTime,omitempty"`                                   // The resolution time recommended by the reporting source of the code issue. (Optional.)
```
rocktavious, line 138: This used to be nullable - but now isn't?

input.go
```diff
@@ -68,586 +68,609 @@ type CategoryDeleteInput struct {
...
+	LevelId               *Nullable[ID]                    `json:"levelId,omitempty" yaml:"levelId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`       // The id of the level the check belongs to. (Optional.)
+	Name                  *Nullable[string]                `json:"name,omitempty" yaml:"name,omitempty" example:"example_value"`                               // The display name of the check. (Optional.)
+	Notes                 *Nullable[string]                `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                             // Additional information about the check. (Optional.)
+	OwnerId               *Nullable[ID]                    `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`       // The id of the owner of the check. (Optional.)
 	UpdateFrequency       *ManualCheckFrequencyUpdateInput `json:"updateFrequency,omitempty" yaml:"updateFrequency,omitempty"`                                 // Defines the minimum frequency of the updates. (Optional.)
```
rocktavious, line 296: This field appears to be optional but isn't wrapped in Nullable?

input.go
```diff
@@ -68,586 +68,609 @@ type CategoryDeleteInput struct {
...
-	Notes                 *string                    `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_notes"`                         // Additional information about the check. (Optional.)
+	Name                  string                     `json:"name" yaml:"name" example:"example_value"`                                               // The display name of the check. (Required.)
+	Notes                 *Nullable[string]          `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                         // Additional information about the check. (Optional.)
+	OwnerId               *Nullable[ID]              `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`   // The id of the team that owns the check. (Optional.)
 	UpdateFrequency       *ManualCheckFrequencyInput `json:"updateFrequency,omitempty" yaml:"updateFrequency,omitempty"`                             // Defines the minimum frequency of the updates. (Optional.)
```
rocktavious, line 281: This field appears to be optional but isn't wrapped in Nullable?

check_test.go
```diff
@@ -1161,6 +1170,9 @@ func TestJsonUnmarshalCreateCheckManual(t *testing.T) {
 	// Act
 	buf1, err1 := ol.UnmarshalCheckCreateInput(ol.CheckTypeManual, []byte(data))
 	// Assert
+	stuff := buf1.(*ol.CheckManualCreateInput)
+	stuff2 := &output
+	fmt.Println(stuff2, stuff)
```
rocktavious, line 1175:
```suggestion
```

scalar_test.go
```diff
@@ -224,7 +224,7 @@ func TestNewIdentifierArray(t *testing.T) {
 }
 
 func TestNewNullString(t *testing.T) {
-	buf, err := json.Marshal(ol.NewNull[string]())
```
rocktavious, line 227: I think the changes in this file can be undone

tags.go
```diff
@@ -222,9 +222,10 @@ func (client *Client) ReconcileTags(resourceType TaggableResourceInterface, tags
 
 	toCreate, toDelete := reconcileTags(tagConnection.Nodes, tagsDesired)
 	for _, tag := range toCreate {
+		taggableResourceType := resourceType.ResourceType()
```
rocktavious, line 225: Its unclear to me why this is changing.

user_test.go
```diff
@@ -38,9 +38,10 @@ func TestInviteUserSkipSendInvite(t *testing.T) {
 	)
 
 	client := BestTestClient(t, "user/invite_skip_send_invite", testRequest)
+	teamMemberRole := ol.UserRoleTeamMember
```
rocktavious, line 41: missed?

rocktavious (approved):
🚀

PR #509, opened 2025-01-27
Code Gen v2

rocktavious:
Resolves #

### Problem

Prior to this a large chunk of the opslevel-go codebase was hand curated.  This is untenniable long term especially when graphql is a schema based API.

### Solution

Finish out the code gen so the entire client library is generated. (or at least as much as possible)

### Checklist

  - If so what is the ticket or PR # (TBD)


actions.go
```diff
@@ -100,7 +78,7 @@ func (client *Client) GetCustomAction(input string) (*CustomActionsExternalActio
 		"input": *NewIdentifier(input),
 	}
 	err := client.Query(&q, v, WithName("ExternalActionGet"))
-	if q.Account.Action.Id == "" {
+	if q.Account.Action.CustomActionsId.Id == "" {
```
rocktavious, line 81: This needs to be investigated to see if we can fix this

PR #511, opened 2025-01-28
Add component CRUD functions

rocktavious:
Resolves #

### Problem

Its confusing to use the CRUD methods for service to create components

### Solution

We want to make it a little nicer for end users to know there are CRUD components but in the end they just federate out to the old services APIs for now.  This will give us a future handle to redirect them to newer APIs as needed.

### Checklist

### DIFF
--- a/service.go
+++ b/service.go
@@ -310,10 +310,7 @@ func (service *Service) GetDocuments(client *Client, variables *PayloadVariables
 
 func (client *Client) CreateService(input ServiceCreateInput) (*Service, error) {
 	var m struct {
-		Payload struct {
-			Service Service
-			Errors  []Error
-		} `graphql:"serviceCreate(input: $input)"`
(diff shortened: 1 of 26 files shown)


service.go
```diff
@@ -376,6 +376,14 @@ func (client *Client) GetService(id ID) (*Service, error) {
 	return &q.Account.Service, nil
 }
 
+func (client *Client) GetService(identifier string) (*Service, error) {
```
rocktavious, line 379: we've been meaning to do this for a LONG time and with the 2025 release emmient - now is a good time to pull the trigger on this breaking change.

For opslevel-go using customers they won't automatically upgrade to this version so it won't break them because the v2025 release is like a "major" version change.  So they'll have to explicitly upgrade and even then the go compiler should catch most of these changes

p25marti (approved):

PR #512, opened 2025-01-29
Codegen Payload Types

rocktavious:
Resolves #

### Problem

Payload types are hand crafted per mutation

### Solution

codegenerate them

### Checklist

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -56,7 +56,7 @@ type CustomActionsTriggerDefinitionsConnection struct {
 
 func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
 	var m struct {
-		Payload struct {
+		Payload struct { // TODO: fix this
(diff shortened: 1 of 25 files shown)


actions.go
```diff
@@ -56,7 +56,7 @@ type CustomActionsTriggerDefinitionsConnection struct {
 
 func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
 	var m struct {
-		Payload struct {
+		Payload struct { // TODO: fix this
```
rocktavious, line 59: All the "TODO" are there for future self's when we come back around.  Unfortunately switching these inline structs to the generated payload stucts requires test changes that i'm not willing to take on right now.  So i've just noted all the places where we need to fix for the future.

p25marti (approved):
LGTM! Went through each of the payloads myself just to be doubly sure.

PR #514, opened 2025-01-30
Add definition for LockedStatus

p25marti:
### Problem
Adds `LockedStatus` to the `PropertyDefinition` struct so that we can use it downstream in our tooling.

### Checklist


rocktavious (approved):
Great Work!  Sorry for all the getting started bumps

PR #515, opened 2025-01-31
Fixes for terraform component types

rocktavious:
Resolves #

### Problem

Connection types were not setup by codegen so this was missed

### Solution

Add connection type for property def's on component types

### Checklist

### DIFF
--- a/component.go
+++ b/component.go
@@ -1,5 +1,7 @@
 package opslevel
 
+import "fmt"
+
 type Component Service
 
 type ComponentCreateInput ServiceCreateInput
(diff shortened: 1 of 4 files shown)


component.go
```diff
@@ -14,6 +16,40 @@ type ComponentTypeConnection struct {
 	TotalCount int             `json:"totalCount" graphql:"-"`
 }
 
+func (s *ComponentType) GetProperties(client *Client, v *PayloadVariables) (*PropertyDefinitionConnection, error) {
```
rocktavious, line 19: This is the pattern we use for connection types because it limits how much data we pull back when doing "list" api calls.

object.go
```diff
@@ -80,12 +80,13 @@ type ComponentTypeId struct {
...
-	IsDefault   bool              // Whether or not the component type is the default (Required)
-	Name        string            // The name of the component type (Required)
-	Timestamps  Timestamps        // When the component type was created and updated (Required)
+	Description string                        // The description of the component type (Optional)
+	Href        string                        // The relative path to link to the component type (Required)
+	Icon        ComponentTypeIcon             // The icon associated with the component type (Required)
+	IsDefault   bool                          // Whether or not the component type is the default (Required)
+	Name        string                        // The name of the component type (Required)
+	Timestamps  Timestamps                    // When the component type was created and updated (Required)
+	Properties  *PropertyDefinitionConnection `graphql:"-"`
```
rocktavious, line 89: This will get to get ported into codegen but this effectively tells the graphql client to ingore this field since we'll populate it with `GetProperties` when called

property.go
```diff
@@ -13,7 +13,7 @@ type PropertyDefinition struct {
 	DisplayType           PropertyDefinitionDisplayTypeEnum `graphql:"displayType" json:"displayType"`
 	PropertyDisplayStatus PropertyDisplayStatusEnum         `graphql:"propertyDisplayStatus" json:"propertyDisplayStatus"`
 	LockedStatus          PropertyLockedStatusEnum          `graphql:"lockedStatus" json:"lockedStatus"`
-	Schema                JSON                              `json:"schema" scalar:"true"`
+	Schema                JSONSchema                        `json:"schema" scalar:"true"`
```
rocktavious, line 16: Not sure why it was coded like this but in our API its actually this type (this struct isn't under codegen yet)

p25marti (approved):
LGTM! Tested it locally 👍

PR #524, opened 2025-03-25
Latest Actions Updates

rocktavious:
Resolves #

### Problem

opslevel-go is not up to date with the API spec

### Solution

Run the code generator

### Checklist


input.go
```diff
@@ -33,6 +33,13 @@ type AliasDeleteInput struct {
 	OwnerType AliasOwnerTypeEnum `json:"ownerType" yaml:"ownerType" example:"domain"` // The resource the alias you wish to delete belongs to (Required)
 }
 
+// ApprovalConfigInput Config for approval
+type ApprovalConfigInput struct {
+	ApprovalRequired *Nullable[bool]        `json:"approvalRequired,omitempty" yaml:"approvalRequired,omitempty" example:"false"` // Flag indicating approval is required (Optional)
```
rocktavious, line 38: This is wrong - it doesn't support json `null`

input.go
```diff
@@ -707,6 +714,7 @@ type ContactUpdateInput struct {
 type CustomActionsTriggerDefinitionCreateInput struct {
 	AccessControl          *CustomActionsTriggerDefinitionAccessControlEnum `json:"accessControl,omitempty" yaml:"accessControl,omitempty" example:"admins"`                          // The set of users that should be able to use the trigger definition (Optional)
 	ActionId               *Nullable[ID]                                    `json:"actionId,omitempty" yaml:"actionId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`           // The action that will be triggered by the Trigger Definition (Optional)
+	ApprovalConfig         *ApprovalConfigInput                             `json:"approvalConfig,omitempty" yaml:"approvalConfig,omitempty"`                                         // Config for approval of action (Optional)
```
rocktavious, line 717: This does support null but @BigBlue79 says it doesn't do what i think it does.

input.go
```diff
@@ -737,6 +746,7 @@ type CustomActionsTriggerDefinitionUpdateInput struct {
 
 // CustomActionsWebhookActionCreateInput Specifies the input fields used in the `customActionsWebhookActionCreate` mutation
 type CustomActionsWebhookActionCreateInput struct {
+	Async          bool                        `json:"async" yaml:"async" example:"false"`                                                                                                                             // Whether the action expects an additional, asynchronous response upon completion (Required)
```
rocktavious, line 749: This isn't breaking because it has the default of false when not present

rocktavious: I think that means it needs a be a pointer in golang so we have 3 states
- not present
- false
- true

input.go
```diff
@@ -33,6 +33,13 @@ type AliasDeleteInput struct {
 	OwnerType AliasOwnerTypeEnum `json:"ownerType" yaml:"ownerType" example:"domain"` // The resource the alias you wish to delete belongs to (Required)
 }
 
+// ApprovalConfigInput Config for approval
+type ApprovalConfigInput struct {
+	ApprovalRequired *bool                  `json:"approvalRequired,omitempty" yaml:"approvalRequired,omitempty" example:"false"` // Flag indicating approval is required (Optional)
+	Teams            *[]IdentifierInput     `json:"teams,omitempty" yaml:"teams,omitempty" example:"[]"`                          // Teams that can approve (Optional)
+	Users            *[]UserIdentifierInput `json:"users,omitempty" yaml:"users,omitempty" example:"[]"`                          // Users that can approve (Optional)
```
rocktavious, line 40: I think these can be just slices - no need for pointers here

input.go
```diff
@@ -737,6 +746,7 @@ type CustomActionsTriggerDefinitionUpdateInput struct {
 
 // CustomActionsWebhookActionCreateInput Specifies the input fields used in the `customActionsWebhookActionCreate` mutation
 type CustomActionsWebhookActionCreateInput struct {
+	Async          *bool                       `json:"async,omitempty" yaml:"async,omitempty" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Required)
```
eapache-opslevel, line 749: the "(Required)" should probably say "(Optional, defaults to `false`)"... we should be able to get that out of the generator tbh, the default value is present in the schema

rocktavious: Ok i'll see if i can add this

input.go
```diff
@@ -747,6 +757,7 @@ type CustomActionsWebhookActionCreateInput struct {
 
 // CustomActionsWebhookActionUpdateInput Inputs that specify the details of a Webhook Action you wish to update
 type CustomActionsWebhookActionUpdateInput struct {
+	Async          *Nullable[bool]              `json:"async,omitempty" yaml:"async,omitempty" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Optional)
```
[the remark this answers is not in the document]
eapache-opslevel: So, per our slack discussion, this should just be `*bool`, right?

input.go
rocktavious: Added

input.go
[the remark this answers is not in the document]
rocktavious: Fixed

object.go
```diff
@@ -23,6 +23,21 @@ type AlertSourceService struct {
...
+// ApprovalConfig Config for approval
+type ApprovalConfig struct {
+	ApprovalRequired bool     // Flag indicating approval is required (Required)
+	Teams            []TeamId // Teams that can approve (Required)
+	Users            []UserId // Users that can approve (Required)
+}
+
+// ApprovalResult The result details of an approvable resource being approved or denied
+type ApprovalResult struct {
+	Actor     Approver             // The actor that performed the approval/denial (Optional)
```
eapache-opslevel, line 35: This is going the other way (output, not input), but should this be `*Approver` if it's optional?

rocktavious: I don't think so - because its a union type, the struct will be initialized on response deserialization but the end user will have to check if the User (or whatever other types we add to the union) will need to be check for validity first before use.  That would be the case if it were a pointer too - they'd just be checking of `result.Actor != nil`  -  I can see how that might be better though rather then having to do `result.Actor.User.Id != ""` and how that would get more complicated once the union grows more types.

With the `EntityOwner` union (which also currently only has a single type - Team) it looks like we did non-pointer too.

Thoughts on the best way to proceed given this context?

eapache-opslevel: So the typical thing for a client to do here is to inject/query `__typename` on the top level of union/interface types, i.e.
```
{
  actor {
    __typename
    ... on User { whatever }
}
```
and then use that to resolve the underlying type into whatever form of polymorphic type system the client language supports (in golang I would assume an empty interface type for unions?). This also makes the distinction between null/non-null much more distinct regardless of the underlying concrete type / whether you queried any fields on that concrete type.

But this is all maybe a bit of a refactor for now. To the more narrow question: it is important for the caller to be able to distinguish the case "the field returned null" from the case "the field returned an unexpected concrete type", which AFAICT requires this (and I guess EntityOwner?) be made pointers that can thus be null if the field returns null, the same as we'd do for other optional object types.

input.go
```diff
@@ -33,6 +33,13 @@ type AliasDeleteInput struct {
 	OwnerType AliasOwnerTypeEnum `json:"ownerType" yaml:"ownerType" example:"domain"` // The resource the alias you wish to delete belongs to (Required)
 }
 
+// ApprovalConfigInput Config for approval
+type ApprovalConfigInput struct {
+	ApprovalRequired *bool                 `json:"approvalRequired,omitempty" yaml:"approvalRequired,omitempty" example:"false"` // Flag indicating approval is required (Optional)
+	Teams            []IdentifierInput     `json:"teams,omitempty" yaml:"teams,omitempty" example:"[]"`                          // Teams that can approve (Optional)
```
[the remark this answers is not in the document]
rocktavious: @eapache-opslevel - ya this has been the struggle with opslevel-go the entire time which is why we are moving to codegen so we can "codify" the rules for things.

So are you saying that `{"teams": null}` is valid input here?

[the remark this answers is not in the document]
rocktavious: @eapache-opslevel ok well if `{"teams": null}` is valid input them switching the `Nullable[T]` is likely the best short term option but i think the real long term one is to make arrays pointers and then use `omitzero` rather then `omitempty` 🤷  - but Nullable is pretty easy to work with and more apprent.

For now i've wrapped it in Nullable.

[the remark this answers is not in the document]
eapache-opslevel: > So are you saying that {"teams": null} is valid input here?

No! Sorry I didn't quite get to answer in time.

I'm saying that omitting the key is importantly different from specifying `{"teams": []}` and based on my understanding of Go, the current version of `[]IdentifierInput` with `omitempty` will omit the key entirely when trying to specify `[]`.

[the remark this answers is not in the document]
rocktavious: ahhh i see

input.go
```diff
@@ -33,6 +33,13 @@ type AliasDeleteInput struct {
 	OwnerType AliasOwnerTypeEnum `json:"ownerType" yaml:"ownerType" example:"domain"` // The resource the alias you wish to delete belongs to (Required)
 }
 
+// ApprovalConfigInput Config for approval
+type ApprovalConfigInput struct {
+	ApprovalRequired *bool                            `json:"approvalRequired,omitempty" yaml:"approvalRequired,omitempty" example:"false"` // Flag indicating approval is required (Optional)
```
andrewstillv15, line 38: the pointer notation indicates this is nullable, correct?

By my (inexperienced) eye, wouldn't we want ApprovalRequired bool? I believe all of `ApprovalRequired`,  `Teams` & `Users`  do not permit a null value.

input.go
```diff
@@ -747,6 +757,7 @@ type CustomActionsWebhookActionCreateInput struct {
 
 // CustomActionsWebhookActionUpdateInput Inputs that specify the details of a Webhook Action you wish to update
 type CustomActionsWebhookActionUpdateInput struct {
+	Async          *bool                        `json:"async,omitempty" yaml:"async,omitempty" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Optional)
```
andrewstillv15, line 760:
```suggestion
	Async          bool                        `json:"async" yaml:"async" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Optional)
```

input.go
```diff
@@ -737,6 +746,7 @@ type CustomActionsTriggerDefinitionUpdateInput struct {
 
 // CustomActionsWebhookActionCreateInput Specifies the input fields used in the `customActionsWebhookActionCreate` mutation
 type CustomActionsWebhookActionCreateInput struct {
+	Async          *bool                       `json:"async,omitempty" yaml:"async,omitempty" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Required Default: false)
```
andrewstillv15, line 749:
```suggestion
	Async          bool                       `json:"async" yaml:"async,omitempty" example:"false"`                                                                                                         // Whether the action expects an additional, asynchronous response upon completion (Required Default: false)
```

object.go
```diff
@@ -23,6 +23,21 @@ type AlertSourceService struct {
...
+type ApprovalConfig struct {
+	ApprovalRequired bool     // Flag indicating approval is required (Required)
+	Teams            []TeamId // Teams that can approve (Required)
+	Users            []UserId // Users that can approve (Required)
+}
+
+// ApprovalResult The result details of an approvable resource being approved or denied
+type ApprovalResult struct {
+	Actor     *Approver            // The actor that performed the approval/denial (Optional)
+	Comment   string               // The comment associated with the approval/denial (Optional)
```
andrewstillv15, line 36: Should this be `Comment   *string`? The comment is nullable in the graphql schema

input.go
eapache-opslevel: We do need to be able to omit the value though?

input.go
eapache-opslevel: We do need to be able to omit the value though?

object.go
eapache-opslevel: 🤔 that is a good question, I think so?

rocktavious: This is on a response type so `null` doesn't really matter for primative types.  Bool's yes but string means it'll just default to `""` which is fine IMO

eapache-opslevel: 👎 probably not this case, but there may be cases where present-but-`""` is importantly different from `null`.

rocktavious: yes - there are cases where Terraform would complain about that but this one won't be part of Terraform.  If you feel strongly I can adjust it.

eapache-opslevel: > there are cases where Terraform would complain about that

Does terraform insist on treating `""` like `null` no matter what?

rocktavious: no it treats them different

input.go
rocktavious: I think this is correct as is

object.go
rocktavious: @andrewstillv15 and I talked IRL and we decided to keep as it is

eapache-opslevel: > no it treats them different

Then what cases would Terraform complain about??

eapache-opslevel: I don't care about this particular case, I don't think it matters for `Comment` particularly, but I do believe strongly that this is the wrong pattern and will probably bite us for some other field where the difference matters.

input.go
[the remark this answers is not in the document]
rocktavious: Fixed and added a test to ensure it

andrewstillv15 (approved):

object.go
rocktavious: `Then what cases would Terraform complain about??`

So generally we get `null` when the customer omits the optional field from their terraform configuration meaning they don't want terraform to manage this field (it can be changed via the UI or other means)  So in that case where something should be "unmanaged" means it shouldn't get written into the state but if it does get written into the state a `""` then it will complain that the customer set it to `null` (by omission) but it returned as `""` (and considers that a bug)

Some of this is on us to properly handle in the Read function and some of it can be patched over by not using `optional` fields in the schema.

I agree is the wrong pattern long term but i don't think it will bite us for this one.  I have about 45 changes i still need to reconcile so i want to get this PR merged and out so @andrewstillv15 can move forward and we can change the default pattern later and reconcile it later.

Issue #496, opened 2024-12-23
Add "Resource Not Found" errors in opslevel-go


rocktavious:
In opslevel-go, add default "Resource Not Found" errors for instances where the API call did not error out but also the returned value is `nil`

PR #525, opened 2025-03-27
Add typed error codes for better downstream client usage of error handling

rocktavious:
Resolves #

### Problem

https://github.com/OpsLevel/opslevel-go/issues/496

Downstream tools that use opslevel-go have no good way to detect errors of certain types easily to perform different logic.  This creates a bunch of boilerplate code in these downstream libraries.

### Solution

Leverage golang's ability to use `errors.As` and `errors.Is` by wrapping the errors and creating special error codes for different common problems to make error handling in downstream code easier and less boilerplate.

This PR only implements it in a handful of places to test the pattern out.  Once its proved working well a followup PR will come that propagates it everywhere.

Additionally this standardization will help future codegen efforts.

### Checklist

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -1,6 +1,7 @@
 package opslevel
 
 import (
+	"errors"
 	"fmt"
 )
 
@@ -51,7 +52,7 @@ type CustomActionsExternalActionsConnection struct {
 type CustomActionsTriggerDefinitionsConnection struct {
 	Nodes      []CustomActionsTriggerDefinition
 	PageInfo   PageInfo
-	TotalCount int
+	TotalCount int `graphql:"-"`
 }
 
 func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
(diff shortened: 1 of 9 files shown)


actions.go
```diff
@@ -51,7 +52,7 @@ type CustomActionsExternalActionsConnection struct {
 type CustomActionsTriggerDefinitionsConnection struct {
 	Nodes      []CustomActionsTriggerDefinition
 	PageInfo   PageInfo
-	TotalCount int
+	TotalCount int `graphql:"-"`
```
rocktavious, line 55: This was found when digging through code - its a pattern we've switched to so just cleaning it up here

errors_test.go
```diff
@@ -1,38 +1,51 @@
...
+	output := ol.HasAPIErrors(errs)
 	// Assert
+	autopilot.Equals(t, true, ol.ErrIs(output, ol.ErrorAPIError))
 	autopilot.Equals(t, `OpsLevel API Errors:
 	- '' can't be blank
 	- '' is not a valid input`,
 		output.Error())
 }
+
+func TestIsResourceFoundError(t *testing.T) {
```
rocktavious, line 43: This is the main goal - we wanted an easy way in downstream tools to beable to detect this kind of error cleanly and consistently.  It will reduce 100's lines of error checking across the tools.

errors.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func IsResourceFound(resource any) error {
+	// TODO: Also Check if ID is valid somehow `.Id == ""`
+	if resource == nil {
+		return NewClientError(ErrorResourceNotFound, fmt.Errorf("resource '%T' not found", resource))
+	}
+	if casted, ok := resource.(Identifiable); ok && casted.GetID() == "" {
+		return NewClientError(ErrorResourceNotFound, fmt.Errorf("resource '%T' not found", resource))
+	}
```
rocktavious, line 72: This is the boilerplate error checking we want to remove from downstream tools.

jasonopslevel (approved):
Still mostly a rubber stamp at this point, but the refactor looks reasonable :)

PR #530, opened 2025-04-21
Allow setting context without changing every method sig

eapache-opslevel:
### Problem

It was not possible to control the `Context` that was used for requests generated by all of the helper methods, e.g. `GetTeam()`.

### Solution

Add client methods to set and reset the default context on the client object. This is not as nice as properly plumbing an argument through to all the helper methods, but that would be a huge painful migration. This does the job for now.

### DIFF
--- a/clientGQL.go
+++ b/clientGQL.go
@@ -11,8 +11,9 @@ import (
 )
 
 type Client struct {
-	pageSize int
-	client   *graphql.Client
+	pageSize       int
+	client         *graphql.Client
+	defaultContext context.Context
 }
 
 func NewGQLClient(options ...Option) *Client {
@@ -40,8 +41,9 @@ func NewGQLClient(options ...Option) *Client {
 		})
 
 	return &Client{
-		pageSize: settings.pageSize,
-		client:   graphql.NewClient(url, standardClient).WithRequestModifier(modifier),
+		pageSize:       settings.pageSize,
+		client:         graphql.NewClient(url, standardClient).WithRequestModifier(modifier),
+		defaultContext: context.Background(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


clientGQL.go
```diff
@@ -60,24 +62,32 @@ func (client *Client) InitialPageVariablesPointer() *PayloadVariables {
 	return &v
 }
 
+func (client *Client) SetCtx(ctx context.Context) {
+	client.defaultContext = ctx
```
rocktavious, line 66: 🤔 would this work properly / break when the client struct is shared via goroutines?  By default we use `*Client ` everywhere so i'm a bit afraid it will break Terraform.  But i guess it only breaks when you use `SetCtx` multiple times.

rocktavious: https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol#terraform-plan

This appears to be the best documentation they have on it and it seems like it calls "ConfigureProvider" every time so that means it should call https://github.com/OpsLevel/terraform-provider-opslevel/blob/main/opslevel/provider.go#L130 every time which should be safe.

rocktavious: I think the true test would be pulling this change into the provider repo using the submodule and running the integration test suite to make sure large parallel operations still work correctly.

I can try doing this tomorrow.

eapache-opslevel: This will 100% break if used for parallel operations. But... you don't have to call it, it's completely optional? So just don't call it in parallel cases like terraform.

That said, maybe it means we should just bite the bullet and plumb it all the way through to every helper method, and deal with the breaking change.

rocktavious: I think because you add a single context to the client function thats what breaks it for parallel cases like terraform weather terraform calls `SetCtx` or not.

FWIW - i did try an integration test in terraform and it failed spectacularly
https://github.com/OpsLevel/terraform-provider-opslevel/actions/runs/14595017030

PR #531, opened 2025-04-22
Kr/dry up connections

rocktavious:
Resolves #

### Problem

Dealing with serializing the "nodes" of a connection type is painful and not easy in downstream tools

### Solution

Give connection types an interface they can implement so we can more easily serialize the "nodes" of a connection type.

### Checklist

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -49,10 +49,18 @@ type CustomActionsExternalActionsConnection struct {
 	TotalCount int
 }
 
+func (s *CustomActionsExternalActionsConnection) GetNodes() any {
+	return s.Nodes
+}
+
 type CustomActionsTriggerDefinitionsConnection struct {
 	Nodes      []CustomActionsTriggerDefinition
 	PageInfo   PageInfo
 	TotalCount int `graphql:"-"`
+}
+
+func (s *CustomActionsTriggerDefinitionsConnection) GetNodes() any {
+	return s.Nodes
 }
 
 func (client *Client) CreateWebhookAction(input CustomActionsWebhookActionCreateInput) (*CustomActionsExternalAction, error) {
--- a/common.go
+++ b/common.go
@@ -89,3 +89,7 @@ func extractAliases(existingAliases, aliasesWanted []string) ([]string, []string
 	}
 	return aliasesToCreate, aliasesToDelete
 }
+
+type Connection interface {
+	GetNodes() any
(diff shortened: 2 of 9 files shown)


input.go
```diff
@@ -936,6 +936,7 @@ type ServiceCreateInput struct {
 	OwnerInput            *IdentifierInput `json:"ownerInput,omitempty" yaml:"owner,omitempty"`                                            // The owner for this service. (Optional.)
 	LifecycleAlias        *string          `json:"lifecycleAlias,omitempty" yaml:"lifecycle,omitempty" example:"example_alias"`            // The lifecycle stage of the service. (Optional.)
 	SkipAliasesValidation *bool            `json:"skipAliasesValidation,omitempty" yaml:"skipAliasesValidation,omitempty" example:"false"` // Allows for the creation of a service with invalid aliases. (Optional.)
+	Type                  *IdentifierInput `json:"type,omitempty" yaml:"type,omitempty"`                                                   // The type of the component. (Optional.)
```
rocktavious, line 939: not sure whats going on here. I need to figure out where this came from

eapache-opslevel, line 939: Unrelated change picked up by the generator?

eapache-opslevel:
I gather that this is just manually-added methods and not a change to the generator?

eapache-opslevel:
_If_ we're generating these it probably makes sense to expand the interface, and generate methods for PageInfo and TotalCount as well. But if they're just manual for now, it's not worth the effort I don't think.

input.go
rocktavious: Ahh - i think i know why this is happening.  The CLI was based off a older version of opslevel-go because of the 2025 nullability change that the CLI is not ready for yet (but thats not the case for mcp since it only uses "list" functions).  So opslevel-mcp's submodule is also based off that.  Now that opslevel-mcp is separate we can bring it up to 2025 and I can cherry pick this against that.

I made this branch from the submodule of opslevel-mcp.

So the order of operations i need to do now
- make a PR to opslevel-mcp to update opslevel-go to latest main branch (should be safe now that its isolated from the CLI)
- cherry pick this to the latest main
- re-review

rocktavious:
@eapache-opslevel yes thats a good idea for when we generate - was faster and easier to just block this out by hand to see if it'll work properly.

PR #532, opened 2025-04-23
Add support for search query on docs

andrewstillv15:
Resolves #

### Problem

Add support for search query on docs

### Checklist


document.go
```diff
@@ -27,3 +33,21 @@ func (client *Client) ServiceApiDocSettingsUpdate(service string, docPath string
 	err := client.Mutate(&m, v, WithName("ServiceApiDocSettingsUpdate"))
 	return &m.Payload.Service, HandleErrors(err, m.Payload.Errors)
 }
+
+type DocumentsConnection struct {
+	Nodes []ServiceDocument
```
rocktavious, line 38: I think we are missing page info here

document.go
```diff
@@ -27,3 +33,21 @@ func (client *Client) ServiceApiDocSettingsUpdate(service string, docPath string
...
+	var q struct {
+		Account struct {
+			Documents DocumentsConnection `graphql:"documents(searchTerm: $searchTerm)"`
+		}
+	}
+
+	v := PayloadVariables{
+		"searchTerm": &searchTerm,
+	}
+	err := client.Query(&q, v, WithName("DocumentSearch"))
```
rocktavious, line 51: This is likely missing automatic pagination (unless we have a restriction of no more then 100 documents on the ruby side)

document.go
```diff
@@ -12,6 +12,12 @@ type ServiceDocumentContent struct {
 	Content string `graphql:"content" json:"content,omitempty"`
 }
 
+type ServiceDocumentsContentConnection struct {
+	Nodes      []ServiceDocumentContent
+	PageInfo   PageInfo
+	TotalCount int `graphql:"-"`
```
eapache-opslevel, line 18: What does the `graphql:"-"` do? I don't see it on other connections.

document.go
eapache-opslevel, line 38: This doesn't seem quite right, as it lacks content? How do we get content from global documents search?

document.go
rocktavious: It means exclude the field from the generated graphql query.  We historcially have taken a stance that 'totalCount' on connection objects is unreliable because sometimes it means total count and sometimes its "filtered count" so the usage on teh ruby side is inconsistent.  So for all paginated connection resources we add the field and then calculate it dynamically based on the final count of nodes returned after all pagination.

document.go
rocktavious: I think we'll actually want 2 functions on the client here.  1 to get the documents list and 1 to get the content for a single document.   Because pulling back all the content for all documents is going to be heavyweight.

document.go
eapache-opslevel: 🤔 isn't that just `len(Nodes)`?

rocktavious: yes but historically we were exposing the field `TotalCount` on the connection type untill we found this problem / bug so the pattern was to keep it but self calculate it.

andrewstillv15:
closed in favour of https://github.com/OpsLevel/opslevel-go/pull/534

PR #534, opened 2025-04-23
Add ListDocs & GetDocs support

andrewstillv15:
Resolves #

### Checklist

### DIFF
--- a/document.go
+++ b/document.go
@@ -1,9 +1,10 @@
 package opslevel
 
 type ServiceDocument struct {
-	Id         ID         `graphql:"id" json:"id"`
-	HtmlURL    string     `graphql:"htmlUrl" json:"htmUrl,omitempty"`
-	Timestamps Timestamps `graphql:"timestamps" json:"timestamps"`
+	Id         ID                    `graphql:"id" json:"id"`
+	HtmlURL    string                `graphql:"htmlUrl" json:"htmUrl,omitempty"`
+	Source     ServiceDocumentSource `graphql:"source" json:"source"`
+	Timestamps Timestamps            `graphql:"timestamps" json:"timestamps"`
 }
 
 type ServiceDocumentContent struct {
--- a/document.go
+++ b/document.go
@@ -48,7 +48,7 @@ func (client *Client) ListDocuments(variables *PayloadVariables) (*ServiceDocume
 	}
 	q.Account.Documents.TotalCount = len(q.Account.Documents.Nodes)
 
-	for q.Account.Documents.PageInfo.HasNextPage {
+	if q.Account.Documents.PageInfo.HasNextPage {
 		(*variables)["after"] = q.Account.Documents.PageInfo.End
 
 		resp, err := client.ListDocuments(variables)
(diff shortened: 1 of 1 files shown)


document.go
```diff
@@ -12,6 +12,12 @@ type ServiceDocumentContent struct {
 	Content string `graphql:"content" json:"content,omitempty"`
 }
 
+type DocumentsConnection struct {
+	Nodes      []ServiceDocument
```
andrewstillv15, line 16: This works well enough for now and I didn't want another identical document struct floating around but i'm open to creating a new Document struct.

eapache-opslevel: There is in fact already a `ServiceDocumentsConnection` with the exact same structure, so I think this is unnecessary?

document.go
```diff
@@ -27,3 +33,48 @@ func (client *Client) ServiceApiDocSettingsUpdate(service string, docPath string
...
+
+	if err := client.Query(&q, *variables, WithName("ListDocuments")); err != nil {
+		return nil, err
+	}
+	q.Account.Documents.TotalCount = len(q.Account.Documents.Nodes)
+
+	for q.Account.Documents.PageInfo.HasNextPage {
+		(*variables)["after"] = q.Account.Documents.PageInfo.End
+
+		resp, err := client.ListDocuments(variables)
```
eapache-opslevel, line 57: This method confusingly mixes recursion (calling itself) with iteration (the `for` loop). On inspection, all of our other paginated methods do the same.

@rocktavious am I missing a subtle reason we need both here, or is this just an original confusion that has been copy-pasta-d throughout?

eapache-opslevel:
I think we also need some way to list the documents for a given service? Whether that is an optional argument to the main `ListDocuments`, or a separate func? Or does that already exist?

document.go
rocktavious: @eapache-opslevel nope you are correct.  Bad pattern propagation from my originally flawed example when i was first getting auto pagination working.

We need to use recursion because you cannot feed the `q` var into `client.Query` multiple times.  So it was originally written with iteration and then fixed to use recursion without undoing the former.

eapache-opslevel:
>  I opted to de-risk of this MR by excluding that work. I'm not opposed to adding it here.

Ah, no if it's coming in a second MR that's totally fine, I was just expecting it here. Keeping this one small is 👌

document.go
```diff
@@ -27,3 +27,51 @@ func (client *Client) ServiceApiDocSettingsUpdate(service string, docPath string
...
+		Account struct {
+			Documents ServiceDocumentsConnection `graphql:"documents(searchTerm: $searchTerm, after: $after, first: $first)"`
+		}
+	}
+
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+
+	if (*variables)["searchTerm"] == nil {
```
eapache-opslevel, line 42: 🤔 the search term shouldn't be required by the server... does leaving it as nil in the variables not work?

eapache-opslevel:
LGTM now

document.go
andrewstillv15: Yea, it complains that the search term is declared but not set

eapache-opslevel: I am liking the hasura client less and less...

rocktavious (approved):

PR #535, opened 2025-04-23
Replace a bunch of confusing/unnecessary loops with ifs

eapache-opslevel:
From https://github.com/OpsLevel/opslevel-go/pull/534#discussion_r2056619141

> We need to use recursion because you cannot feed the q var into client.Query multiple times. So it was originally written with iteration and then fixed to use recursion without undoing the former then copy pasted.

Just find/replace all the `for`s with `if`s to make this a bit clear. There might be more thorough cleanup possible but this is the high-value/low-effort version.

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -30,7 +30,7 @@ func (customActionsTriggerDefinition *CustomActionsTriggerDefinition) ExtendedTe
 		return nil, err
 	}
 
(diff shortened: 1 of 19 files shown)


rocktavious (approved):
LGTM - assuming tests pass.  We have full coverage testing on these

PR #538, opened 2025-04-24
Auto populate pagination params on variables

andrewstillv15:
Resolves #

### Problem

Round out setting pagination logic such that they are truly optional params.

### Checklist

### DIFF
--- a/clientGQL.go
+++ b/clientGQL.go
@@ -100,3 +100,17 @@ func (client *Client) Validate() error {
 func WithName(name string) graphql.Option {
 	return graphql.OperationName(name)
 }
+
+func (client *Client) PopulatePaginationParams(variables *PayloadVariables) *PayloadVariables {
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+
+	if (*variables)["after"] == nil {
+		(*variables)["after"] = client.InitialPageVariables()["after"]
+	}
+	if (*variables)["first"] == nil {
+		(*variables)["first"] = client.InitialPageVariables()["first"]
+	}
+	return variables
+}
--- a/document.go
+++ b/document.go
@@ -35,9 +35,7 @@ func (client *Client) ListDocuments(variables *PayloadVariables) (*ServiceDocume
 		}
 	}
 
-	if variables == nil {
-		variables = client.InitialPageVariablesPointer()
-	}
+	variables = client.PopulatePaginationParams(variables)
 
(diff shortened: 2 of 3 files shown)


clientGQL.go
```diff
@@ -100,3 +100,17 @@ func (client *Client) Validate() error {
 func WithName(name string) graphql.Option {
 	return graphql.OperationName(name)
 }
+
+func (client *Client) PopulatePaginationParams(variables *PayloadVariables) *PayloadVariables {
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+
+	if (*variables)["after"] == nil {
```
andrewstillv15, line 109: There isn't a spread operator in go from what I can tell, very open to other suggestions for this.

rocktavious: I think the function you have it good but for future writing something that merges 2 maps would be more future proof.

```
func mergeMaps(map1, map2 map[string]any) map[string]any {
	merged := make(map[string]any)

	for key, value := range map1 {
		merged[key] = value
	}

	for key, value := range map2 {
		if _, present := merged[key]; !present {
			merged[key] = value
		}
	}

	return merged
}
```

Then you can just have something like

```
func (client *Client) PopulatePaginationParams(variables *PayloadVariables) *PayloadVariables {
	if variables == nil {
	}
	
	return mergeMaps(variables, client.InitialPageVariablesPointer())
}
```

andrewstillv15:
@rocktavious I've hit all the low hanging fruit with respect to the variables == nil behaviour. There are a few (4) more spots with a bit more complexity that I can look into to but before I do, I want to make sure that the testing scope of this isn't much beyond the test cases run in the CI. These changes are non-breaking so it should be safe, but I just want to double check before continuing.

rocktavious:
@andrewstillv15 - Good to stop and check.  I think most of our tests use "nil" on the `client.List*` methods.  You'd have to do a search but if i'm right then you'll want to add a test or 2 that pass in partially filled PayloadVariables to one of the `client.List*` methods on whatever is easiest to implement.

I think we potentially have some non-nil ones in some of the sub resource calls on like Service or Teams maybe.

The last option (and this is likely WAY simplier to implement) is to just put the new common.go functions under test.  As long as `PopulatePaginationParams` has some unit tests baking its expected behaviour into a test case (or a few) then it doesn't matter if our `client.List*` methods are tested with non-nil variables input.

PR #542, opened 2025-05-14
Add default value to documentation on fields based on schema

rocktavious:
Resolves #

### Problem

Certain fields on input structs have defaults

### Solution

the schema tells us the defaults so lets add it to the comments describing the field

### Checklist

### DIFF
--- a/input.go
+++ b/input.go
@@ -82,10 +82,10 @@ type CategoryUpdateInput struct {
 // CheckAlertSourceUsageCreateInput Specifies the input fields used to create an alert source usage check
 type CheckAlertSourceUsageCreateInput struct {
 	AlertSourceNamePredicate *PredicateInput         `json:"alertSourceNamePredicate,omitempty" yaml:"alertSourceNamePredicate,omitempty"`           // The condition that the alert source name should satisfy to be evaluated (Optional)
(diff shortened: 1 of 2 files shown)


input.go
```diff
@@ -82,10 +82,10 @@ type CategoryUpdateInput struct {
 // CheckAlertSourceUsageCreateInput Specifies the input fields used to create an alert source usage check
 type CheckAlertSourceUsageCreateInput struct {
 	AlertSourceNamePredicate *PredicateInput         `json:"alertSourceNamePredicate,omitempty" yaml:"alertSourceNamePredicate,omitempty"`           // The condition that the alert source name should satisfy to be evaluated (Optional)
-	AlertSourceType          *AlertSourceTypeEnum    `json:"alertSourceType,omitempty" yaml:"alertSourceType,omitempty" example:"datadog"`           // The type of the alert source (Optional)
+	AlertSourceType          *AlertSourceTypeEnum    `json:"alertSourceType,omitempty" yaml:"alertSourceType,omitempty" example:"custom"`            // The type of the alert source (Optional)
```
rocktavious, line 85: This is changing because the enum's where changed last time to have a new value in the `[0]` spot of the array so the code gen is updating the example to the `[0]` value in the array.

PR #543, opened 2025-05-14
Add alert source APIs

rocktavious:
Resolves #

### Problem

We have new APIs for working with alert sources more generically but not way to do this via opslevel-go

### Solution

Codegen the structs and add the necessary client functions for this

### Checklist


p25marti (approved):
LGTM!

PR #545, opened 2025-05-14
Add ability to get last deploy of a service

rocktavious:
Resolves #

### Problem

Given a service there is no way to get the last deploy for it.

### Solution

Add a utility function to the service struct to get the last deploy field.  It was done as a add on function so we don't bloat the ListServices call further.

### Checklist

### DIFF
--- a/.changes/unreleased/Feature-20250514-150622.yaml
+++ b/.changes/unreleased/Feature-20250514-150622.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: Add utility function to service struct to `GetLastDeploy` event
+time: 2025-05-14T15:06:22.701659-05:00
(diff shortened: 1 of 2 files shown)


deploys.go
```diff
@@ -0,0 +1,30 @@
...
+	DeployStatus  DeployStatusEnum // The normalized status of the deploy. This is derived from the status field (Optional)
+	DeployUrl     string           // The url the where the deployment can be found (Optional)
+	DeployedAt    iso8601.Time     // The time the deployment happened (Optional)
+	DeployerEmail string           // The email of who is responsible for the deployment (Optional)
+	DeployerName  string           // The name of who is responsible for the deployment (Optional)
+	Description   string           // The given description of the deploy (Required)
+	Environment   string           // The environment in which the deployment happened in (Optional)
+	Id            ID               // The id of the deploy (Required)
+	Service       ServiceId        // The service object the deploy is attached to (Optional)
+	StartedAt     iso8601.Time     // The time the deploy started (Optional)
```
rocktavious, line 22: I need to run this through codegen first because this is not all the fields.

derek-etherton-opslevel (approved):

PR #546, opened 2025-05-14
Kr/deprecations

rocktavious:
Resolves #

### Problem

None

### Solution

During the latest round of client-gen it found a bunch of deprecations

### Checklist


jasonopslevel:
We don't plan on removing the deprecated methods anytime soon, so we could alternative choose to not remove them from opslevel-go either. If there's an option in the codegen to ignore deprecations, we could do that too.

PR #549, opened 2025-05-19
Refactor Connection types with codegen

rocktavious:
Resolves #

### Problem

The connection types in our API are not generated by client-gen and thus everything for them in opslevel-go is hand crafted.

### Solution

Actually generate the most common connection types that use `nodes` (for now excluding ones that truly need edges) and then fix up any none standard code because of this unification

Additionally I took the time to bite off the change to stop using `totalCount` everywhere since its unreliable between different queries (especially ones with filtering on the "list" call)  So this just calculates on the client side the final total based on the final length of the nodes array after all pagination (and nested pagination) is done.

### Checklist

- [X] I have run this code, and it appears to resolve the stated issue.
- [X] This PR has no user interface changes or has already received approval from product management to change the interface.


cache_test.go
```diff
@@ -25,32 +25,32 @@ func TestCache(t *testing.T) {
...
 	testRequestFive := autopilot.NewTestRequest(
-		`query CategoryList($after:String!$first:Int!){account{rubric{categories(after: $after, first: $first){nodes{description,id,name},{{ template "pagination_request" }},totalCount}}}}`,
+		`query CategoryList($after:String!$first:Int!){account{rubric{categories(after: $after, first: $first){nodes{description,id,name},{{ template "pagination_request" }}}}}}`,
 		`{ "after": "", "first": 100 }`,
 		`{"data":{"account":{"rubric":{ "categories":{ "nodes":[{{ template "category_1" }}] } }}}}`,
 	)
 	testRequestSix := autopilot.NewTestRequest(
-		`{account{rubric{levels{nodes{alias,description,id,index,name},{{ template "pagination_request" }},totalCount}}}}`,
-		`{}`,
+		`query LevelsList($after:String!$first:Int!){account{rubric{levels(after: $after, first: $first){nodes{alias,description,id,index,name},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor}}}}}`,
```
rocktavious, line 38:
```suggestion
		`query LevelsList($after:String!$first:Int!){account{rubric{levels(after: $after, first: $first){nodes{alias,description,id,index,name},{{ template "pagination_request" }}}}}}`,
```

level_test.go
```diff
@@ -69,8 +69,8 @@ func TestGetMissingRubricLevel(t *testing.T) {
 func TestListRubricLevels(t *testing.T) {
 	// Arrange
 	testRequest := autopilot.NewTestRequest(
-		`{account{rubric{levels{nodes{alias,description,id,index,name},{{ template "pagination_request" }},totalCount}}}}`,
-		`{}`,
+		`query LevelsList($after:String!$first:Int!){account{rubric{levels(after: $after, first: $first){nodes{alias,description,id,index,name},pageInfo{hasNextPage,hasPreviousPage,startCursor,endCursor}}}}}`,
```
rocktavious, line 72:
```suggestion
		`query LevelsList($after:String!$first:Int!){account{rubric{levels(after: $after, first: $first){nodes{alias,description,id,index,name},{{ template "pagination_request" }}}}}}`,
```

level_test.go
```diff
@@ -114,10 +114,11 @@ func TestListRubricLevels(t *testing.T) {
 	)
 	client := BestTestClient(t, "rubric/level/list", testRequest)
 	// Act
-	result, _ := client.ListLevels()
+	result, err := client.ListLevels(nil)
```
rocktavious, line 117: This is the only customer facing change and since its a argument change in a type safe language customers code (or our other tools) will detect it at compile time

derek-etherton-opslevel (approved):

PR #553, opened 2025-05-22
fix service LastDeploy field to not use pagination variables

derek-etherton-opslevel:
Resolves #

### Problem

Fixes that we were defaulting `GetLastDeploy` params to pagination params, when those aren't accepted for this (singular) field.

### Checklist

### DIFF
--- a/service.go
+++ b/service.go
@@ -229,8 +229,9 @@ func (service *Service) GetLastDeploy(client *Client, variables *PayloadVariable
 	if service.Id == "" {
 		return nil, fmt.Errorf("unable to get LastDeploy, invalid service id: '%s'", service.Id)
 	}
+
 	if variables == nil {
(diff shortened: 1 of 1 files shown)


service.go
```diff
@@ -229,8 +229,9 @@ func (service *Service) GetLastDeploy(client *Client, variables *PayloadVariable
 	if service.Id == "" {
 		return nil, fmt.Errorf("unable to get LastDeploy, invalid service id: '%s'", service.Id)
 	}
+
 	if variables == nil {
-		variables = client.InitialPageVariablesPointer()
+		variables = &PayloadVariables{}
 	}
```
derek-etherton-opslevel, line 235: considered getting rid of the `variables` param altogether, but figure keeping it is:
a. backwards compatible with current version of opslevel-mcp, and
b. more future-proof?

PR #555, opened 2025-05-27
Add ability to “query” for a given relationship

rocktavious:
Resolves #

### Problem

Terraform needs the ability to query for a relationship by ID so that it can detect drift.  This is not currently possible

### Solution

Make this possible.

### Checklist

### DIFF
--- a/relationship.go
+++ b/relationship.go
@@ -1,5 +1,7 @@
 package opslevel
 
+import "fmt"
+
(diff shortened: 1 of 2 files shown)


rocktavious:
@derek-etherton-opslevel tests are failing because the main branch is broken due to some other fixes i pushed for Terraform.

derek-etherton-opslevel (approved):

PR #559, opened 2025-05-28
Add the user name to the lightweight UserId struct

rocktavious:
Resolves #

### Problem

The CLI command `opslevel list team -o json` outputs a `memberships` section that uses `UserId` which currently only has `Id` and `Email` for the user but a customer would like to also have `Name` which is entirely reasonable.

### Solution

Move the `Name` field to the `UserId` struct fixes this and doesn't cause any "interface" changes to the end user just to the shape of the queries made - user name - should be lightweight enough and expected information when you are getting just the `UserId` struct back so this should be a 100% safe change.

### Checklist


derek-etherton-opslevel (approved):
LGTM

PR #562, opened 2025-06-06
Add ServiceStats to Service

wesleyjellis:
Resolves https://gitlab.com/jklabsinc/OpsLevel/-/work_items/13509

### Problem

We don't support service stats and rubric the API

### Solution

Add those things with the new `Service.GetServiceStats` method. It's a beefy query so I didn't want to fetch it by default.

I did have to add `CheckId` to avoid loops where levels have checks which have levels which have checks which have levels which have checks....

PR to client-gen goes with this too, incoming shortly

### Checklist


rocktavious:
@wesleyjellis ya i'm pretty certain it won't affect terraform - but the CLI does alot of "implict" stuff where if the struct is part of some other struct as a nested field  - now the CLI is missing output data that someone might be depending on.

PR #565, opened 2025-06-11
Add Campaigns

wesleyjellis:
Add Campaign, CampaignConnection and ListCampaigns

### Checklist


campaign.go
```diff
@@ -0,0 +1,77 @@
...
+		defaultAfter := (*defaultPages)["after"].(string)
+		campaignVariables.After = &defaultAfter
+	}
+
+	if campaignVariables.SortBy == nil {
+		campaignVariables.SortBy = &CampaignSortEnumStartDateDesc
+	}
+
+	if campaignVariables.Status == nil {
+		campaignVariables.Status = &CampaignStatusEnumInProgress
```
rocktavious, line 50: If the customer didn't supply "status" and since there is no default value for it - my understanding is that the default case of using this function will fail - IE `client. ListCampaigns(nil)` will result in

```
query CampaignsList {
  account {
    campaigns(
      first: 100
      after: ""
      sortBy: null
      filter: [{key: status, arg: ""}]
    ) {
    ```
    
    and this fails for me with
    
    ```
    {
  "errors": [
    {
      "message": "invalid filter_params",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "account",
        "campaigns"
      ]
    }
  ],
  "data": null
}
    ```

rocktavious (approved):
LGTM

PR #570, opened 2025-06-29
Allow client transport to be overridden

### LINKED ISSUE #197: Allow setting a custom transport when instantiating clients

samstarling:
This change fixes #197, and is based on #263.

### Problem

In our codebase, we use `*httpmock.MockTransport` to mock third parties. However, this requires being able to override the transport used by the OpsLevel client, which is currently not possible.

### Solution

Introduces `SetTransport` as a client option, and ensures this is applied to both the GraphQL and REST clients. In #263, it was mentioned that support should be added for both.

### Checklist

- [ ] Does this change require a Terraform schema change? **No, it doesn't**

### DIFF
--- a/.changes/unreleased/Feature-20250629-123543.yaml
+++ b/.changes/unreleased/Feature-20250629-123543.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: Allow the HTTP transport of GraphQL and REST clients to be overridden
+time: 2025-06-29T12:35:43.236187+01:00
--- a/client.go
+++ b/client.go
@@ -2,6 +2,7 @@ package opslevel
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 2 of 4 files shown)


samstarling:
cc @rocktavious and @zakisk, because I know you were involved in this last time

derek-etherton-opslevel (approved):
Looks great, thank you!

@rocktavious flagged that there may be a bit of cruftiness with how the `retryablehttp` client is also wrapping the Transport - let us know if that causes you any issues. We should be able to upgrade [go-graphql-client](https://github.com/hasura/go-graphql-client/releases/tag/v0.14.0) soon to help clean things up a bit.

rocktavious:
@samstarling and @lawrencejones  - This is now in opslevel-go version [2025.7.28](https://github.com/OpsLevel/opslevel-go/releases/tag/v2025.7.28)

PR #571, opened 2025-07-03
Add Contacts to User

derek-etherton-opslevel:
### Problem

The User struct does not have `Contacts`.

### Solution

Add `Contacts` to the User struct.

Unit tests should be good, but manually verified w/ this:

```go
func main() {
	apiToken := os.Getenv("OPSLEVEL_API_TOKEN")
	if apiToken == "" {
		log.Fatal("OPSLEVEL_API_TOKEN environment variable is not set")
	}

	client := opslevel.NewGQLClient(opslevel.SetAPIToken(apiToken))

	users, err := client.ListUsers(nil)
	if err != nil {
		log.Fatalf("Failed to list users: %v", err)
	}

	for _, user := range users.Nodes {
		fmt.Printf("User: %s\n", user.Name)
		fmt.Printf("Contacts: %+v\n", user.Contacts)
	}
}
```

### Checklist

### DIFF
--- a/object.go
+++ b/object.go
@@ -613,6 +613,7 @@ type UserId struct {
 // User A user is someone who belongs to an organization
 type User struct {
 	UserId
+	Contacts      []Contact         // The contacts for the user (Optional)
 	HtmlUrl       string            // A link to the HTML page for the resource. Ex. https://app.opslevel.com/services/shopping_cart (Required)
 	ProvisionedBy ProvisionedByEnum // What provisioned this user (Optional)
 	Role          UserRole          // The user's assigned role (Optional)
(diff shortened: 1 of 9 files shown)


object.go
```diff
@@ -613,6 +613,7 @@ type UserId struct {
 // User A user is someone who belongs to an organization
 type User struct {
 	UserId
+	Contacts      []Contact         // The contacts for the user (Optional)
```
derek-etherton-opslevel, line 616: this is the only actual change - rest is test "fixtures".

 I contemplated cutting this into a new `UserWithContacts` struct so it wasn't included by default on e.g. Team Managers, but didn't see any issues with scale when we include it by default:

- No user in prod has >4 contacts
- There's no n+1 when requesting a list of users (one `contacts` query)
- Users is pretty lightweight today (unlike Components)

jasonopslevel: yeah, should be fine.

jasonopslevel (approved):
lgtm

PR #575, opened 2025-07-29
Latest updates from API Schema

rocktavious:
Resolves #

### Problem

The client hasn't been generated from the API schema in over 2 months.

### Solution

Perform that generation and only commit what makes sense to change.

### Checklist

### DIFF
--- a/input.go
+++ b/input.go
@@ -380,7 +380,7 @@ type CheckRelationshipUpdateInput struct {
 	Notes                      *string                 `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                                                         // Additional information about the check (Optional)
 	OwnerId                    *Nullable[ID]           `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                   // The id of the owner of the check (Optional)
 	RelationshipCountPredicate *PredicateInput         `json:"relationshipCountPredicate,omitempty" yaml:"relationshipCountPredicate,omitempty"`                                       // The condition that should be satisfied by the number of RelatedTo relationships (Optional)
-	RelationshipDefinitionId   *Nullable[ID]           `json:"relationshipDefinitionId,omitempty" yaml:"relationshipDefinitionId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // Count relationships of a specific relationship definition (Optional)
+	RelationshipDefinitionId   *ID           `json:"relationshipDefinitionId,omitempty" yaml:"relationshipDefinitionId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // Count relationships of a specific relationship definition (Optional)
 }
 
 // CheckRepositoryFileCreateInput Specifies the input fields used to create a repo file check
--- a/input.go
+++ b/input.go
@@ -1160,12 +1160,12 @@ type ServiceDependencyKey struct {
 
 // ServiceFilterInput Input to be used to filter types
 type ServiceFilterInput struct {
-	Arg           *Nullable[string]    `json:"arg,omitempty" yaml:"arg,omitempty" example:"example_value"`             // Value to be filtered (Optional)
-	CaseSensitive *Nullable[bool]      `json:"caseSensitive,omitempty" yaml:"caseSensitive,omitempty" example:"false"` // Whether or not the filter should be case sensitive (Optional)
-	Connective    *ConnectiveEnum      `json:"connective,omitempty" yaml:"connective,omitempty" example:"and"`         // The logical operator to be used in conjunction with multiple filters (requires predicates to be supplied) (Optional Default: or)
-	Key           *ServiceFilterEnum   `json:"key,omitempty" yaml:"key,omitempty" example:"alert_status"`              // Field to be filtered (Optional)
-	Predicates    []ServiceFilterInput `json:"predicates,omitempty" yaml:"predicates,omitempty" example:"[]"`          // A list of service filter input (Optional)
-	Type          *TypeEnum            `json:"type,omitempty" yaml:"type,omitempty" example:"belongs_to"`              // Type of operation to be applied to value on the field (Optional Default: equals)
+	Arg           string               `json:"arg,omitempty" yaml:"arg,omitempty" example:"example_value"`     // Value to be filtered (Optional)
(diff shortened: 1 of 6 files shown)


input.go
```diff
@@ -354,6 +354,35 @@ type CheckPackageVersionUpdateInput struct {
...
+	LevelId                    ID                      `json:"levelId" yaml:"levelId" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                   // The id of the level the check belongs to (Required)
+	Name                       string                  `json:"name" yaml:"name" example:"example_value"`                                                           // The display name of the check (Required)
+	Notes                      *string                 `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                                     // Additional information about the check (Optional)
+	OwnerId                    *Nullable[ID]           `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`               // The id of the team that owns the check (Optional)
+	RelationshipCountPredicate PredicateInput          `json:"relationshipCountPredicate" yaml:"relationshipCountPredicate"`                                       // The condition that should be satisfied by the number of RelatedTo relationships (Required)
```
rocktavious, line 367: need to validate if this is required

input.go
```diff
@@ -657,12 +686,67 @@ type CheckToolUsageUpdateInput struct {
...
+	IntroducedAt  *Nullable[iso8601.Time]               `json:"introducedAt,omitempty" yaml:"introducedAt,omitempty" example:"2025-01-05T01:00:00.000Z"` // Timestamp when this issue was introduced (Optional)
+	IssueCategory *Nullable[string]                     `json:"issueCategory,omitempty" yaml:"issueCategory,omitempty" example:"example_value"`          // Category of this code issue. Required to create code issue (Optional)
+	Name          *Nullable[string]                     `json:"name,omitempty" yaml:"name,omitempty" example:"example_value"`                            // Name of the code issue. Required to create code issue (Optional)
+	Severity      *Nullable[string]                     `json:"severity,omitempty" yaml:"severity,omitempty" example:"example_value"`                    // Severity level of this issue (Optional)
+	Url           *Nullable[string]                     `json:"url,omitempty" yaml:"url,omitempty" example:"example_value"`                              // URL with more information about this issue (Optional)
```
rocktavious, line 704: These need to be vetted - pointers might suffice - not sure if `null` is needed for any of these.

input.go
rocktavious: This is good

input.go
```diff
@@ -354,6 +354,35 @@ type CheckPackageVersionUpdateInput struct {
...
+	FilterId                   *Nullable[ID]           `json:"filterId,omitempty" yaml:"filterId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                 // The id of the filter the check belongs to (Optional)
+	Id                         ID                      `json:"id" yaml:"id" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                                                 // The id of the check to be updated (Required)
+	LevelId                    *Nullable[ID]           `json:"levelId,omitempty" yaml:"levelId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                   // The id of the level the check belongs to (Optional)
+	Name                       *Nullable[string]       `json:"name,omitempty" yaml:"name,omitempty" example:"example_value"`                                                           // The display name of the check (Optional)
+	Notes                      *string                 `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                                                         // Additional information about the check (Optional)
```
derek-etherton-opslevel, line 380:
```suggestion
	Notes                      *Nullable[string]                 `json:"notes,omitempty" yaml:"notes,omitempty" example:"example_value"`                                                         // Additional information about the check (Optional)
```

input.go
```diff
@@ -354,6 +354,35 @@ type CheckPackageVersionUpdateInput struct {
...
+	OwnerId                    *Nullable[ID]           `json:"ownerId,omitempty" yaml:"ownerId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"`                                   // The id of the owner of the check (Optional)
+	RelationshipCountPredicate *PredicateInput         `json:"relationshipCountPredicate,omitempty" yaml:"relationshipCountPredicate,omitempty"`                                       // The condition that should be satisfied by the number of RelatedTo relationships (Optional)
```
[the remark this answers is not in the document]
derek-etherton-opslevel: same question can be applied to `LevelId`, `CategoryId`, and `RelationshipDefinitionId`. All four of these fields technically accept `null` (nothing blows up), but it does nothing. Functionally, these fields can't be assigned to null.

derek-etherton-opslevel (approved):
LGTM! Some nits on nullability; I'm not sure whether allow it or not is our current practice, you can educate me here.

I re-tested the service filter methods with a couple of my complex use cases and it all continues to work correctly 👍

input.go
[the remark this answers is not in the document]
rocktavious: `LevelId`, `CategoryId` are just following the same setup as the other check input types so i wouldn't want to change those unless we change all check types.

The reason i didn't make `RelationshipCountPredicate` and `RelationshipDefinitionId` nullable is that null doesn't make sense in the update context because they are "required" fields in Create so there is no "null = unset" case during update - there is just omission and thats what a pointer does when `omitempty` is set.

So IMO - nothing to change here.

input.go
rocktavious: This is following the same pattern as the other check types.  If this is "incorrect" then we'll need to address it across all check types with a wider breaking change.  So for now i'm skipping this because we don't want this check type to work differently then the other check types right now.

input.go
```diff
@@ -354,6 +354,35 @@ type CheckPackageVersionUpdateInput struct {
...
+	RelationshipDefinitionId   *Nullable[ID]           `json:"relationshipDefinitionId,omitempty" yaml:"relationshipDefinitionId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // Count relationships of a specific relationship definition (Optional)
```
rocktavious, line 383:
```suggestion
	RelationshipDefinitionId   *ID           `json:"relationshipDefinitionId,omitempty" yaml:"relationshipDefinitionId,omitempty" example:"Z2lkOi8vc2VydmljZS8xMjM0NTY3ODk"` // Count relationships of a specific relationship definition (Optional)
```

PR #613, opened 2026-04-22
Add category field to ComponentType and ComponentTypeInput

saditya370:
## Summary
- Adds `Category` to `ComponentTypeInput` (write) and `ComponentType` (read), backing the recently-public `category` field on `componentTypeCreate` / `componentTypeUpdate` and the `ComponentType` type.
- Valid values: `default`, `infrastructure`, `people`. Setting `infrastructure` surfaces the component type in the Infrastructure catalog.
- Updates the GraphQL selection set and test fixtures.

## Test plan
- [x] `go test ./...` passes
- Create/Update mutation tests send `category` and assert it on the response
- List test asserts `category` is populated on all three fixtures

### DIFF
--- a/.changes/unreleased/component-type-category.yaml
+++ b/.changes/unreleased/component-type-category.yaml
@@ -1,2 +1,2 @@
 kind: Feature
-body: Add `Category` field to `ComponentType` and `ComponentTypeInput` for creating and reading component type categories (`default`, `infrastructure`, `people`)
+body: Add `Category` field to `ComponentType` and `ComponentTypeInput` for creating and reading component type categories (`default`, `infrastructure`). Gated on the backend behind the `infra_component_types_ui` feature flag — accounts without the flag will see an empty `Category` and cannot send the field on input.
--- a/.changes/unreleased/component-type-category.yaml
+++ b/.changes/unreleased/component-type-category.yaml
@@ -1,0 +1,2 @@
+kind: Feature
+body: Add `Category` field to `ComponentType` and `ComponentTypeInput` for creating and reading component type categories (`default`, `infrastructure`, `people`)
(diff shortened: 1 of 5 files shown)


wesleyjellis (commented):
2 major questions:

1. How does this work without the feature flag? I'm pretty sure infrastructure component types are gated behind a feature flag, and I'm not sure what happens if you try to make one without the feature flag

2. Do we need the people category? It looks like we don't let you make people componenttypes inside opslevel, so I think we should drop it from any tests, at least as far as creating them goes... Is that going to cause a problem if you try to manage your componenttypes with Terraform though? ie: terraform tries to delete all the component types it doesn't know about?

saditya370:
> 2 major questions:
> 
> 1. How does this work without the feature flag? I'm pretty sure infrastructure component types are gated behind a feature flag, and I'm not sure what happens if you try to make one without the feature flag
> 2. Do we need the people category? It looks like we don't let you make people componenttypes inside opslevel, so I think we should drop it from any tests, at least as far as creating them goes... Is that going to cause a problem if you try to manage your componenttypes with Terraform though? ie: terraform tries to delete all the component types it doesn't know about?

Flag: Backend gates category behind infra_component_types_ui on both input and output. Without the flag the field is just absent from the schema, so the SDK sends nothing (omitempty) and reads come back as "". Flagless accounts are unaffected. Added a note to the schema docs + changelog.

people: Dropped. Validator is now ("default", "infrastructure") and the test fixtures don't reference people anymore. No drift risk for existing people-category types — the provider only touches what's in the config, it doesn't enumerate-and-delete unmanaged types. Only edge case is if someone imports a people type into state, which would hit the validator.

PR #614, opened 2026-05-04
Add tags and teams to User struct and hydrate them in ListUsers

saditya370:
## Summary

Adds `Tags *TagConnection` and `Teams *TeamIdConnection` to the `User` struct so `ListUsers` returns tag and team-membership data inline, mirroring the precedent on `Service`. Without this, callers (notably the OpsLevel MCP server) cannot answer questions like "which team does this user belong to?" or "which users are tagged as on-call leads?" without a follow-up query per user.

### `object.go`
`User` gains two pointer-typed connection fields:
- `Tags *TagConnection` (`json:"tags,omitempty"`)
- `Teams *TeamIdConnection` (`json:"teams,omitempty"`)

These match the pattern used on `Service.Tags`. The graphql library auto-generates the inline selection set from the struct, so `ListUsers` now requests `tags{nodes{id,key,value},pageInfo}` and `teams{nodes{alias,id},pageInfo}` per user node.

### `user.go`
- Renamed `User.Teams(...)` → `User.GetTeams(...)` to free the `Teams` field name. The previous method name collided with the new struct field. Aligns naming with `UserId.GetTags`.
- New `User.Hydrate(client)` follows pagination on the embedded `Tags` and `Teams` connections when their inner `pageInfo.hasNextPage` is `true`. Mirrors `Service.Hydrate`.
- `ListUsers` now invokes `Hydrate` on **every** node from the initial query — not just nodes merged from recursive list pages. This closes a silent-truncation bug: previously, if every user fit on the first outer page (the common case for small accounts), no user ever had its inner tag/team pagination followed, capping users at the schema-default 500 nodes per inner connection without a follow-up.

#### `user_test.go`
- Updated 7 hardcoded query-string assertions to include the new `tags`/`teams` selections under each user node.
- New `TestListUserPopulatesTagsAndTeams` — asserts the inline path works: a user returned from `ListUsers` with embedded tags+teams JSON gets those fields populated on the Go struct.
- New `TestListUserHydratesFirstPageInnerConnections` — regression for the first-page hydration fix. Sets `tags.pageInfo.hasNextPage=true` on a user returned in the first outer page and asserts a follow-up `UserTagsList` query is issued and the merged tag list is returned.

#### `team_test.go`, `cache_test.go`, `actions_test.go`, `clientGQL_test.go`
Every test that selects `User` as a sub-field (`Team.Manager`, `deletedMembers`) now expects `tags{...}` and `teams{...}` inside the `manager{...}` (or equivalent) selection. Same pattern in all four files.

#### `testdata/templates/query/team/get.tpl`
The shared `team_get` template's `manager{...}` block was extended with `tags` and `teams` connection selections to match the new `User` shape.

## Notes

- `TeamIdConnection` (not `TeamConnection`) is used for the inline selection — minimal `{alias, id}` per team. Team `name` is intentionally not added; most callers can use the alias as a stable identifier.
- The unparameterized `tags{...}` / `teams{...}` selections rely on the schema's `default_page_size: 100` / `default_max_page_size: 500`. `Hydrate` follows `hasNextPage` for both, so power users with overflowing tags/teams are still fully populated.

## Test plan

- [x] `go test ./...` passes
- [x] `TestListUserPopulatesTagsAndTeams` covers the inline path
- [x] `TestListUserHydratesFirstPageInnerConnections` covers the first-page hydration fix
- [x] All existing user/team/cache/actions/clientGQL test query strings updated and passing


user.go
```diff
@@ -61,7 +61,7 @@ func (userId *UserId) GetTags(client *Client, variables *PayloadVariables) (*Tag
 	return &q.Account.User.Tags, nil
 }
 
-func (user *User) Teams(client *Client, variables *PayloadVariables) (*TeamIdConnection, error) {
+func (user *User) GetTeams(client *Client, variables *PayloadVariables) (*TeamIdConnection, error) {
```
andrewstillv15: Can we do this without changing this method name? We'll have update the major version we change this.

saditya370: @andrewstillv15   Renamed the struct field to TeamsConnection with an explicit graphql:"teams" tag instead, which keeps User.Teams intact

user.go
```diff
@@ -92,6 +92,40 @@ func (user *User) Teams(client *Client, variables *PayloadVariables) (*TeamIdCon
 	return &q.Account.User.Teams, nil
 }
 
+func (user *User) Hydrate(client *Client) error {
```
andrewstillv15, line 95: Nit

```suggestion
func (user *User) hydrate(client *Client) error {
```

andrewstillv15 (approved):

PR #624, opened 2026-06-29
Add RunnerJob.InitCommands/InitImage and RunnerJobVariable.Scope

derek-etherton-opslevel:
Adds two new fields for configuring the `initContainer` of a RunnerJob: `RunnerJob.InitCommands/InitImage`

Also adds a new `scope` field to `RunnerJobVariable` with possible values of `""`, `"init"`, or `"main"`, allowing the partitioning of variables to the init or main container. `""`, the default value, will just include variables in both containers.

`opslevel-runner` changes supporting these additions to follow

Pre-requisite for: https://github.com/OpsLevel/opslevel-runner/pull/304

### DIFF
--- a/job.go
+++ b/job.go
@@ -72,13 +72,13 @@ type Runner struct {
 }
 
 // RunnerJobVariableScope controls which container(s) on the job pod receive a
-// variable. An empty value means the variable is available in both the startup
-// init container and the main job container (the default).
+// variable. An empty value means the variable is available in both the init
+// container and the main job container (the default).
 type RunnerJobVariableScope string
 
 const (
-	RunnerJobVariableScopeStartup RunnerJobVariableScope = "startup"
-	RunnerJobVariableScopeMain    RunnerJobVariableScope = "main"
+	RunnerJobVariableScopeInit RunnerJobVariableScope = "init"
+	RunnerJobVariableScopeMain RunnerJobVariableScope = "main"
 )
 
 type RunnerJobVariable struct {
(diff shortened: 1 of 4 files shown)


job.go
```diff
@@ -83,13 +94,15 @@ type RunnerJobFile struct {
...
-	Files     []RunnerJobFile      `json:"files"`
+	Commands     []string             `json:"commands"`
+	Id           ID                   `json:"id"`
+	Image        string               `json:"image"`
+	Outcome      RunnerJobOutcomeEnum `json:"outcome"`
+	Status       RunnerJobStatusEnum  `json:"status"`
+	Variables    []RunnerJobVariable  `json:"variables"`
+	Files        []RunnerJobFile      `json:"files"`
+	InitCommands []string             `json:"initCommands"`
+	InitImage    string               `json:"initImage"`
```
derek-etherton-opslevel, line 105: `initImage` is not technically required at present, but could be handy in short order

job.go
```diff
@@ -71,10 +71,21 @@ type Runner struct {
...
 }
 
+// RunnerJobVariableScope controls which container(s) on the job pod receive a
+// variable. An empty value means the variable is available in both the init
+// container and the main job container (the default).
+type RunnerJobVariableScope string
+
+const (
+	RunnerJobVariableScopeInit RunnerJobVariableScope = "init"
+	RunnerJobVariableScopeMain RunnerJobVariableScope = "main"
```
jasonopslevel, line 81: presumably we could also extend it with `sidecar` or something too!

jasonopslevel (approved):

PR #625, opened 2026-07-08
Bump Go to 1.25 and x/crypto to v0.52.0, x/net to v0.55.0 (8 CVEs)

derek-etherton-opslevel:
This PR raises the minimum Go version and patches eight critical CVEs across `golang.org/x/crypto/ssh` and `golang.org/x/net/idna`.

**Downstream projects that import `opslevel-go` must now be built with Go 1.25.0 or newer** (up from 1.24). This is because `golang.org/x/crypto` v0.52.0 sets `go 1.25.0` as its minimum in its own `go.mod`, and Go picks the highest `go` directive across the module graph.

CI reads the Go version from `go.mod` (`go-version-file: 'go.mod'` in every workflow), so no workflow changes are needed for this repo.

## Go requirement

- `go.mod` `go` directive: `1.24` → `1.25.0`
- `toolchain go1.24.1` line removed (implied by the new directive)

### `golang.org/x/crypto` 0.41.0 → 0.52.0

Commit history between tags: https://github.com/golang/crypto/compare/v0.41.0...v0.52.0

Patches **7 critical CVEs**, all in `golang.org/x/crypto/ssh`:

- [GO-2026-5020](https://pkg.go.dev/vuln/GO-2026-5020) When writing data larger than 4GB in a single `Write` call on an SSH channel, an integer overflow in the internal payload size calculation caused the write loop to spin indefinitely, sending empty packets without making progress. The size comparison now uses `int64` to prevent truncation.
- [GO-2026-5023](https://pkg.go.dev/vuln/GO-2026-5023) Previously, CVE-2024-45337 fixed an authorization bypass for misused ssh server configurations; if any other type of callback is passed other than public key, then the source-address validation would be skipped.
- [GO-2026-5006](https://pkg.go.dev/vuln/GO-2026-5006) When adding a key to a remote agent, constraint extensions such as `restrict-destination-v00@openssh.com` were not serialized in the request. Destination restrictions were silently stripped when forwarding keys, allowing unrestricted use of the key on the remote host. The client now serializes all constraint extensions. Additionally, the in-memory keyring returned by `NewKeyring()` now rejects keys with unsupported constraint extensions instead of silently ignoring them.
- [GO-2026-5017](https://pkg.go.dev/vuln/GO-2026-5017) A malicious SSH peer could send unsolicited global request responses to fill an internal buffer, blocking the connection's read loop. The blocked goroutine could not be released by calling `Close()`, resulting in a resource leak per connection. Unsolicited global responses are now discarded.
- [GO-2026-5019](https://pkg.go.dev/vuln/GO-2026-5019) The `Verify()` method for FIDO/U2F security key types (`sk-ecdsa-sha2-nistp256@openssh.com`, `sk-ssh-ed25519@openssh.com`) did not check the User Presence flag. Signatures generated without physical touch were accepted, allowing unattended use of a hardware security key. To restore the previous behavior, return a `no-touch-required` extension in `Permissions.Extensions` from `PublicKeyCallback`.
- [GO-2026-5021](https://pkg.go.dev/vuln/GO-2026-5021) Previously, a revoked `SignatureKey` belonging to a CA was not correctly checked for revocation. Now, both the `key` and `key.SignatureKey` are checked for `@revoked`.
- [GO-2026-5005](https://pkg.go.dev/vuln/GO-2026-5005) The in-memory keyring returned by `NewKeyring()` silently accepted keys with the `ConfirmBeforeUse` constraint but never enforced it. The key would sign without any confirmation prompt, with no indication to the caller that the constraint was not in effect. `NewKeyring()` now returns an error when unsupported constraints are requested.

### `golang.org/x/net` 0.43.0 → 0.55.0

Commit history between tags: https://github.com/golang/net/compare/v0.43.0...v0.55.0

Patches **1 critical CVE** in `golang.org/x/net/idna`:

- [GO-2026-5026](https://pkg.go.dev/vuln/GO-2026-5026) The `ToASCII` and `ToUnicode` functions incorrectly accept Punycode-encoded labels that decode to an ASCII-only label. For example, `ToUnicode("xn--example-.com")` incorrectly returned the name `example.com`, potentially bypassing name-based allowlists or masking the true label. This can lead to privilege escalation in programs performing checks on the ASCII hostname.

## Other changes

`go mod tidy` also pulled these sibling `golang.org/x/*` modules forward:
- `golang.org/x/mod` v0.26.0 → v0.35.0
- `golang.org/x/sync` v0.16.0 → v0.20.0
- `golang.org/x/sys` v0.35.0 → v0.45.0
- `golang.org/x/term` v0.34.0 → v0.43.0
- `golang.org/x/text` v0.28.0 → v0.37.0
- `golang.org/x/tools` v0.35.0 → v0.44.0
- Adds `golang.org/x/tools/go/expect` and `golang.org/x/tools/go/packages/packagestest` as indirect deps (transitively-required test tooling)
- Removes stale `gopkg.in/check.v1` entry

🤖 Generated with [Claude Code](https://claude.com/claude-code)


derek-etherton-opslevel:
In addition to CI, did a quick smoke test on listing users against this branch w/ no issues

```go
	client := opslevel.NewGQLClient(opslevel.SetAPIToken(token))
	users, err := client.ListUsers(nil)
```

> OK — fetched 17 users, total in account: 17

andrewstillv15 (approved):

PR #628, opened 2026-07-17
Add Kubernetes integration create/update with ETL definitions

saditya370:
## Summary

Adds Go client support for managing a Kubernetes infra integration's ETL config
(extract/transform definitions) via the public GraphQL API. Customer ask from
GitLab issue #14148 (Truelayer wants to manage their Kubernetes integration ETL
as code). The schema side is already merged in the monolith (MR !19177); this is
the client counterpart.

## Changes

- `KubernetesIntegrationInput` (`name`, `extractDefinition`, `transformDefinition`
  as `*Nullable[string]`) — hand-written alongside `AWSIntegrationInput` since
  `input.go` is generated.
- `CreateIntegrationKubernetes(input)` → `kubernetesIntegrationCreate`
- `UpdateIntegrationKubernetes(identifier, input)` → `kubernetesIntegrationUpdate`
- Read path: `KubernetesIntegrationFragment` (`ExtractDefinition`,
  `TransformDefinition`) added to the `Integration` struct as
  `... on KubernetesIntegration`, with the shared `integration_request` test
  template updated in the matching position so existing recorded queries still match.
- Tests: `TestCreateKubernetesIntegration` / `TestUpdateKubernetesIntegration`
  assert the ETL definitions round-trip through the mutations.
- Changie entry under `.changes/unreleased/`.

Mutation/input/field names verified against the monolith's committed public
schema dump (`public/schemas/schema.graphql`).

## Testing

- `go build ./...`
- `go test -race ./...` — full suite passes (existing integration tests double as
  a regression check for the shared query-template change)
- `task lint` (gofumpt + golangci-lint) — 0 issues

### DIFF
--- a/integration.go
+++ b/integration.go
@@ -57,9 +57,9 @@ type AWSIntegrationInput struct {
 }
 
 type KubernetesIntegrationInput struct {
-	ExtractDefinition   *Nullable[string] `json:"extractDefinition,omitempty"`
+	ExtractDefinition   *Nullable[YAML]   `json:"extractDefinition,omitempty"`
 	Name                *Nullable[string] `json:"name,omitempty"`
-	TransformDefinition *Nullable[string] `json:"transformDefinition,omitempty"`
+	TransformDefinition *Nullable[YAML]   `json:"transformDefinition,omitempty"`
 }
 
 func (awsIntegrationInput AWSIntegrationInput) GetGraphQLType() string { return "AwsIntegrationInput" }
--- a/integration.go
+++ b/integration.go
@@ -37,6 +37,11 @@ type GoogleCloudIntegrationFragment struct {
 	TagsOverrideOwnership bool                 `graphql:"tagsOverrideOwnership"`
 }
 
+type KubernetesIntegrationFragment struct {
+	ExtractDefinition   string `graphql:"extractDefinition"`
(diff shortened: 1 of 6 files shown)


integration.go
```diff
@@ -51,7 +56,17 @@ type AWSIntegrationInput struct {
 	RegionOverride       *[]string         `json:"regionOverride,omitempty"`
 }
 
+type KubernetesIntegrationInput struct {
+	ExtractDefinition   *Nullable[YAML]   `json:"extractDefinition,omitempty"`
```
andrewstillv15:
```suggestion
	ExtractDefinition   *YAML   `json:"extractDefinition,omitempty"`
```

integration.go
```diff
@@ -51,7 +56,17 @@ type AWSIntegrationInput struct {
...
+	Name                *Nullable[string] `json:"name,omitempty"`
+	TransformDefinition *Nullable[YAML]   `json:"transformDefinition,omitempty"`
```
andrewstillv15:
```suggestion
	TransformDefinition *YAML   `json:"transformDefinition,omitempty"`
```

andrewstillv15: We'll have to run the tidy task as well after these

andrewstillv15 (approved):

PR #631, opened 2026-08-20
Add custom integration create/update with ETL definitions

saditya370:
## Summary

Adds opslevel-go support for managing Custom Integrations, including their
Extract/Transform mapping configuration:

- `CustomIntegrationFragment` — exposes `extractDefinition` / `transformDefinition` on reads
- `CustomIntegrationETLInput` — the full create/update input
- `CreateIntegrationCustom` → `customIntegrationCreate`
- `UpdateIntegrationCustom` → `customIntegrationUpdate`

Deletes already work via the existing generic `DeleteIntegration`.

This mirrors the Kubernetes integration pair added in #628. It unblocks a
matching `opslevel_integration_custom` resource in terraform-provider-opslevel
(OPS-387), where the customer ask originated.

## Why the input type isn't called `CustomIntegrationInput`

`input.go` already declares a generated `CustomIntegrationInput`, but it only
has `name` — client-gen appears to drop YAML-scalar fields, so the generated
type can't express the ETL definitions and is referenced nowhere. Since
`input.go` is regenerated (most recently in #630), editing it would be
clobbered, and the name is taken.

So the usable input is hand-maintained in `integration.go` as
`CustomIntegrationETLInput`, with `GetGraphQLType()` mapping it back onto the
schema's `CustomIntegrationInput` — the same pattern `AWSIntegrationInput`
already uses for the `AwsIntegrationInput` mismatch. `KubernetesIntegrationInput`
is likewise hand-written for the same reason.

Worth a follow-up with whoever owns client-gen: the stale generated stub is
dead, exported API surface.

## Notes for reviewers

- **`testdata/templates/integrations.tpl`** — adding a fragment to `Integration`
  changes the generated query, so the shared `integration_request` template
  needed the matching `... on CustomIntegration{...}`. Every integration test
  asserts against it.
- **Kubernetes test churn** — `Integration` now embeds two fragments with
  `ExtractDefinition`/`TransformDefinition`, making the bare selector ambiguous.
  The four assertions now use `result.KubernetesIntegrationFragment.…`.
  No behavior change. (Azure/GoogleCloud already share field names this way.)
- **`customSyncInterval` is deliberately excluded** — it's a non-public,
  feature-flagged argument on the API and shouldn't be in a public client.

## Testing

`go build`, `go vet`, `gofumpt`, `golangci-lint` (0 issues), and `go test ./...`
all pass. New: `TestCreateCustomIntegration`, `TestUpdateCustomIntegration`.

### DIFF
--- a/integration.go
+++ b/integration.go
@@ -29,6 +29,11 @@ type AzureResourcesIntegrationFragment struct {
 	TenantId              string   `graphql:"tenantId"`
 }
 
+type CustomIntegrationFragment struct {
+	ExtractDefinition   string `graphql:"extractDefinition"`
+	TransformDefinition string `graphql:"transformDefinition"`
+}
+
 type GoogleCloudIntegrationFragment struct {
 	Aliases               []string             `graphql:"aliases"`
 	ClientEmail           string               `graphql:"clientEmail"`
@@ -56,6 +61,15 @@ type AWSIntegrationInput struct {
 	RegionOverride       *[]string         `json:"regionOverride,omitempty"`
 }
 
+// CustomIntegrationInput in the generated input.go omits the YAML-scalar ETL
+// fields, so the full input is hand-maintained here, as for KubernetesIntegrationInput.
+// GetGraphQLType maps it back onto the schema's CustomIntegrationInput.
+type CustomIntegrationETLInput struct {
+	ExtractDefinition   *YAML             `json:"extractDefinition,omitempty"`
+	Name                *Nullable[string] `json:"name,omitempty"`
+	TransformDefinition *YAML             `json:"transformDefinition,omitempty"`
+}
+
 type KubernetesIntegrationInput struct {
 	ExtractDefinition   *YAML             `json:"extractDefinition,omitempty"`
 	Name                *Nullable[string] `json:"name,omitempty"`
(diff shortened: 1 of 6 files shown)


integration.go
```diff
@@ -56,13 +61,26 @@ type AWSIntegrationInput struct {
 	RegionOverride       *[]string         `json:"regionOverride,omitempty"`
 }
 
+// CustomIntegrationInput in the generated input.go omits the YAML-scalar ETL
+// fields, so the full input is hand-maintained here, as for KubernetesIntegrationInput.
+// GetGraphQLType maps it back onto the schema's CustomIntegrationInput.
+type CustomIntegrationETLInput struct {
```
Farjaad: why are we calling it ETL input and not just input?

saditya370: due to naming collision: generated input.go already defines CustomIntegrationInput, but it only has name and omits extractDefinition/transformDefinition. Since input.go is regenerated, editing that type directly would be clobbered on the next regen, and Go wont let us define another CustomIntegrationInput in this package.

I renamed the hand-written type from CustomIntegrationETLInput to CustomIntegrationMappingInput and kept GetGraphQLType() returning "CustomIntegrationInput" so it still maps to the GraphQL schema type.

Farjaad: why can't we update the existing input? these fields are additive/optional so I thought it would be okay. I've never worked on this repo so that's why I am asking

Farjaad: if we add more fields to `CustomIntegrationInput`, I wouldn't want to create yet another type in opslevel-go

saditya370: Updated to use the existing CustomIntegrationInput directly. I initially avoided input.go because it is generated, but agreed that since these fields are optional/additive on the same GraphQL input, this is cleaner than adding a second type.

Farjaad (approved):

PR #642, opened 2026-08-24
Fix CustomIntegrationInput.Name type, add generated KubernetesIntegrationInput

andrewstillv15:
## Summary
- `CustomIntegrationInput.Name` is now `*string` instead of `*Nullable[string]`. The backend declares `validates: { allow_null: false, allow_blank: false }` on this field, so an explicit `null` is rejected outright — confirmed against the backend's own `"fails if name is null"` test in `custom_integration_create_test.rb`.
- Adds the generated `KubernetesIntegrationInput` to `input.go`, replacing the hand-maintained duplicate that previously lived in `integration.go` (client-gen had never generated this type because the `YAML` scalar had no example value, so it silently produced nothing for it — see companion fix in `client-gen`). `KubernetesIntegrationInput.Name` stays `*Nullable[string]`: it has no `allow_null` validator, and sending `null` was verified (via `kubernetesIntegrationCreate`/`kubernetesIntegrationUpdate` mutation tests) to be silently reset to a computed default name rather than rejected.
- Updates `integration_test.go` to construct `CustomIntegrationInput.Name` with `NewString` instead of `RefOf`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` — all passing, including `TestCreateCustomIntegration`, `TestUpdateCustomIntegration`, `TestCreateKubernetesIntegration`, `TestUpdateKubernetesIntegration`
- [x] Verified both field-nullability behaviors empirically against the backend's own test suite (not just inferred from schema declarations)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/input.go
+++ b/input.go
@@ -894,9 +894,9 @@ type CustomActionsWebhookActionUpdateInput struct {
 
 // CustomIntegrationInput Input for upserting a custom integration
 type CustomIntegrationInput struct {
-	ExtractDefinition   *YAML             `json:"extractDefinition,omitempty" yaml:"extractDefinition,omitempty" example:"example_value"`     // The configured definition for extracting data from inbound webhooks or HTTP polling (Optional)
-	Name                *Nullable[string] `json:"name,omitempty" yaml:"name,omitempty" example:"example_value"`                               // Name of the custom integration type (Optional)
-	TransformDefinition *YAML             `json:"transformDefinition,omitempty" yaml:"transformDefinition,omitempty" example:"example_value"` // The configured definition for transforming extracted data from inbound webhooks or HTTP polling to another resource (Optional)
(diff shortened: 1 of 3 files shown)


input.go
```diff
@@ -1001,6 +1001,13 @@ type InfrastructureResourceSchemaInput struct {
 	Type string `json:"type" yaml:"type" example:"example_value"` // The type of the infrastructure resource (Required)
 }
 
+// KubernetesIntegrationInput Specifies the input fields used to create and update a Kubernetes integration
```
andrewstillv15: This was added by hand in the wrong place, I've regenerated the `KubernetesIntegrationInput` & `CustomIntegrationInput` declarations using client-gen.

andrewstillv15:
@Farjaad, I think this should go in before https://github.com/OpsLevel/terraform-provider-opslevel/pull/662 but i don't think it's required

jasonopslevel (approved):

PR #643, opened 2026-08-25
Add ListTeamsIncludingProperties for batched team metadata reads

saditya370:
### Problem

`Team.Properties` is tagged `graphql:"-"`, so it is never selected by generated
queries. Any caller that wants properties for a list of teams has to call
`(*Team).GetProperties` per team.

For the Terraform provider's `opslevel_teams` data source that is one request
per team. Measured against a 263-team account, the per-team approach issued
~266 requests and took **26 seconds**, close enough to the provider's 30s
client timeout that one of two runs died mid-read. It is not viable at
customer scale.

### Solution

`ListTeamsIncludingProperties` selects the properties connection inline
alongside the team list, so listing every team costs **one request per page**
instead of one per team. Same account, same data: **3 requests, under a
second** end to end through the provider.

`TeamWithProperties` embeds `Team` and redeclares `Properties` with a graphql
tag. Go field shadowing means the embedded `graphql:"-"` field is skipped and
ours is selected, so the query asks for `properties` exactly once. That is
subtle enough to be worth a guard, so `TestListTeamsIncludingProperties`
asserts the full constructed query string.

Because the test registers a **single** request and the autopilot harness fails
on any request it was not told to expect, a per-team fan-out regression would
surface as an unregistered call rather than silently passing.

Tags and memberships are completed by the existing `Team.Hydrate`, which issues
no request unless a connection actually spills past its first page. The inlined
properties connection is selected without pagination arguments, so the API
returns its first 100; a team holding more than that is topped up individually.

**Alternatives considered**

- *Remove `graphql:"-"` from `Team.Properties`.* Simplest, but it would make
  every existing Team query heavier for every consumer. Rejected.
- *Implement this in the Terraform provider.* Works, but couples the provider
  to an opslevel-go internal modeling decision and leaves other consumers
  (`opslevel-mcp` calls `ListTeams` too) to reinvent it. Rejected in favour of
  fixing it at the client layer.
- *Naming.* Deliberately **not** `ListTeamsWithProperties` — every existing
  `ListXWith*` in this repo means *filtered by* (`ListTeamsWithManager`,
  `ListServicesWithTier`), so that name would read as "teams that have
  properties".

**No regression by construction:** the diff is 146 insertions and 0 deletions.
Not one existing line changed.

### Noted but not fixed here

`ListTeams` only calls `Hydrate` on nodes from its recursive call, so teams on
the **first** page are never hydrated. With a default page size of 500, that
means most accounts get no hydration at all: `Responsibilities` stays
HTML-escaped and `TotalCount` stays 0. The same pattern appears in 17 places
across 5 files. Filing separately, since fixing it changes observable values
for every existing consumer and does not belong in an additive PR.

### Checklist

  - Yes — OpsLevel/terraform-provider-opslevel PR for OPS-27, which consumes this and must merge after this releases.
- [x] Make a changie entry that explains the customer facing outcome of this change

### DIFF
--- a/team.go
+++ b/team.go
@@ -409,6 +409,77 @@ func (client *Client) ListTeams(variables *PayloadVariables) (*TeamConnection, e
 	return &q.Account.Teams, nil
 }
 
+// TeamWithProperties is a Team with its custom properties already loaded.
+//
+// Team.Properties is excluded from generated queries with `graphql:"-"`, so callers that
+// need properties for many teams would otherwise pay one extra request per team.
+// Embedding Team and redeclaring the field selects the connection inline instead.
+type TeamWithProperties struct {
+	Team
+	Properties PropertiesConnection `graphql:"properties"`
+}
+
+// ListTeamsIncludingProperties returns every team with its tags and properties populated,
+// costing one request per page of teams rather than one request per team.
+//
+// The inlined properties connection is selected without pagination arguments, so the API
+// returns its first 100 entries; any team holding more than that is topped up on its own.
+func (client *Client) ListTeamsIncludingProperties(variables *PayloadVariables) ([]TeamWithProperties, error) {
+	if variables == nil {
+		variables = client.InitialPageVariablesPointer()
+	}
+
+	teams := make([]TeamWithProperties, 0)
+	for {
+		var q struct {
+			Account struct {
+				Teams struct {
+					Nodes    []TeamWithProperties
+					PageInfo PageInfo
+				} `graphql:"teams(after: $after, first: $first)"`
+			}
+		}
+		if err := client.Query(&q, *variables, WithName("TeamListIncludingProperties")); err != nil {
+			return nil, err
+		}
+		teams = append(teams, q.Account.Teams.Nodes...)
+		if !q.Account.Teams.PageInfo.HasNextPage {
+			break
+		}
+		(*variables)["after"] = q.Account.Teams.PageInfo.End
+	}
+
+	for i := range teams {
+		// Hydrate covers the tags and memberships that came back inline; it issues no
+		// request unless one of those connections actually spilled past its first page.
+		if err := teams[i].Hydrate(client); err != nil {
+			return nil, err
+		}
+		if err := teams[i].hydrateProperties(client); err != nil {
+			return nil, err
+		}
+	}
+	return teams, nil
+}
+
+// hydrateProperties collects any pages of properties beyond the first that the list query
+// already returned. It issues no request for a team holding 100 properties or fewer.
+func (team *TeamWithProperties) hydrateProperties(client *Client) error {
+	// Point the embedded Team at the inlined connection so both views agree, and so
+	// GetProperties appends later pages onto the nodes already returned.
+	team.Team.Properties = &team.Properties
+
+	if !team.Properties.PageInfo.HasNextPage {
+		team.Properties.TotalCount = len(team.Properties.Nodes)
+		return nil
+	}
+
+	variables := client.InitialPageVariablesPointer()
+	(*variables)["after"] = team.Properties.PageInfo.End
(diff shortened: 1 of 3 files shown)


team.go
```diff
@@ -409,6 +409,77 @@ func (client *Client) ListTeams(variables *PayloadVariables) (*TeamConnection, e
...
+	Team
+	Properties PropertiesConnection `graphql:"properties"`
+}
+
+// ListTeamsIncludingProperties returns every team with its tags and properties populated,
+// costing one request per page of teams rather than one request per team.
+//
+// The inlined properties connection is selected without pagination arguments, so the API
+// returns its first 100 entries; any team holding more than that is topped up on its own.
+func (client *Client) ListTeamsIncludingProperties(variables *PayloadVariables) ([]TeamWithProperties, error) {
```
andrewstillv15: I don't think that this solution is very composable. If we want to extend what teams returns, we'll have to deprecate this and replace it with something else. I'll come up with a recommendation.

andrewstillv15: I think that instead of doing `ListTeamsIncludingProperties` or hydrating properties per team, we can remove the properties opt out & hydrate them per page instead of per team.

saditya370: Reworked to your suggestion, dropped the opt-out, properties come inline now, and ListTeamsIncludingProperties is gone

One extra thing I pulled in: ListTeams now hydrates each page. Without it "per page" doesn't quite work — properties come back on page one, but TotalCount stays 0 and overflow never gets fetched. Can split it out if you'd rather.

Tested locally: 10 teams, 1 request, properties inline.

team.go
```diff
@@ -26,7 +26,7 @@ type Team struct {
 	ParentTeam       TeamId
 	Responsibilities string
 	Tags             *TagConnection
-	Properties       *PropertiesConnection `graphql:"-" json:"-"`
+	Properties       *PropertiesConnection `json:"-"`
```
andrewstillv15: why do we need `json:"-"`?

team.go
```diff
@@ -103,6 +103,20 @@ func (team *Team) Hydrate(client *Client) error {
...
 	team.Tags.TotalCount = len(team.Tags.Nodes)
+
+	if team.Properties == nil {
+		team.Properties = &PropertiesConnection{}
+	}
+	if team.Properties.PageInfo.HasNextPage {
+		variables := client.InitialPageVariablesPointer()
+		(*variables)["after"] = team.Properties.PageInfo.End
+		// GetProperties appends onto team.Properties, so the pages already returned
+		// inline by the outer query are kept.
```
andrewstillv15:
```suggestion

```

team.go
```diff
@@ -391,18 +405,22 @@ func (client *Client) ListTeams(variables *PayloadVariables) (*TeamConnection, e
 		return nil, err
 	}
 
+	// Hydrate inner connections for every team on this page. Without this, teams on
+	// the first outer page would never paginate their nested connections beyond the
+	// server's default page size.
```
andrewstillv15:
```suggestion
```

team.go
```diff
@@ -391,18 +405,22 @@ func (client *Client) ListTeams(variables *PayloadVariables) (*TeamConnection, e
...
+	for i := range q.Account.Teams.Nodes {
```
andrewstillv15: Good catch!

team.go
saditya370: It was only there because properties were never populated  a permanently-nil field just cluttered the JSON. Now that they come back inline it doesn't earn its keep, so I've dropped it. Tags right above it has no json tag either, so this matches. Only visible change is CLI json/yaml output gains a properties key.

team.go
saditya370: done!

team.go
saditya370: done

andrewstillv15 (approved):

